PR #10, opened 2015-04-30
Missing CSS folder in pattern-lab/core/styleguide. Adjusted css filter i...

fillerwriter:
...n .gitignore to only ignore root css folder


EvanLovely:
Thanks @fillerwriter !

PR #19, opened 2015-06-28
Tell users to start w/grunt

joshkoenig:
Helpful to have an explicit next-step so people don't wonder.


EvanLovely:
Sometimes it's easy to forget those things that seem obvious to you; thanks!

PR #22, opened 2015-10-04
DO NOT MERGE YET. Node pattern lab, config revamp

illepic:
Early work around simplifying config, and using node pattern lab.


EvanLovely:
Since config has moved to a new module, [`p2-theme-core`](https://github.com/phase2/p2-theme-core), this PR can't be merged to this repo. Basically that new module is just the `grunt-tasks/` folder, many changes should be able to be brought over to it.

PR #33, opened 2016-03-07
minor sass updates

tjheffner:
The scripts in footer.html were auto-added, not sure if they actually need to be there or not.

The clearfix div makes the demo homepage stop stacking images on each other so it looks like a functional layout.

Adding a margin to sg-main improves readability when viewing all patterns


EvanLovely:
So much as changed that I doubt we still have this issue. Let me know if it seems like we could still use this.

PR #47, opened 2016-08-07
Remove quotes.

wheatpenny:
With quotes, Firefox inspector gave an error warning, and the underline was present for the default link. With the quotes removed, the link worked as expected (no underline by default, underline present on hover.)


EvanLovely:
thanks!

PR #49, opened 2016-09-07
add render filter to to twig-components

webcultist:


EvanLovely:
Thanks! I sent this code to where it really comes from (though we do commit theme here after installing): https://github.com/pattern-lab/plugin-drupal-twig-components/blob/master/dist/filters/render.filter.php

webcultist:
I originally added the code there as you described in the open issue, but could not get it to work, till I recognized, that the theme gets it's functions from this folder - otherwise I would have added it already there ^^

PR #55, opened 2016-10-02
[HOLD] Introduce the default 'startup:start' and 'startup:stop' tasks

illepic:
These tasks are introduced in this PR to p2-theme-core: https://github.com/phase2/p2-theme-core/pull/15

Essentially, we now have a "startup mode" that certain tasks can check against to allow them to not run redundantly on first `npm run start`. This modifies the `default` task in pattern-lab-starter to add in these two tasks.


illepic:
CI will pass once https://github.com/phase2/p2-theme-core/pull/15 is merged.

EvanLovely:
Fixed in `P2-theme-core` v9.0.3

PR #70, opened 2017-03-04
Fixes #69 - Bower and gulp don't use locally installed versions

### LINKED ISSUE #69: Bower and Gulp don't use versions installed by dependencies
These are declared as dependencies in package.json but unless someone has ./node_modules/.bin in their path they won't be found and used allowing/requiring these to be installed globally.

tekante:


tekante:
Hmm, maybe it's only using the global bower in the post install command. Will do some testing to confirm.

tekante:
Looks like my diagnosis that a global bower was needed was incorrect and the PATH magic is taking effect so I'm going to close this one out as denied.

PR #71, opened 2017-03-08
Add composer.json

bradjones1:
Allows for installing via Composer into the themes/contrib directory. Also adds GPL-2 license, as would be expected for a Drupal project, to address https://github.com/phase2/pattern-lab-starter/issues/40

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,13 @@
+{
+  "name": "phase2/pattern-lab-starter",
+  "type": "drupal-theme",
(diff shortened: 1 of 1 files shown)


bradjones1:
Realized there's a composer.json in the pattern-lab directory, and that brings in the pattern lab static site generator requirements. Still likely needs some re-engineering re: discovery on d.o. and packagist, but this isn't necessary I don't think.

PR #73, opened 2017-03-25
Remove aleksip/plugin-data-transform

tekante:
Remove this as a default due to the memory usage and performance impacts combined with the relative infrequency of use of the feature this affords.


EvanLovely:
Good call; this was helpful at first when older methods were used but not anymore. PS PHP 7 compiles PL about 5 times faster than PHP 5 and with much less memory usage.

PR #75, opened 2017-03-29
[HOLD] Adding Webpack

EvanLovely:
Let's get Webpack in! I've pushed out `p2-theme-core` v9.4 which adds support for it. This branch has disabled the previous JS tasks that aggregate and babelify all JS files in the theme into one big JS file and have currently set up the very basics of some JS files in `./js`. There are several things left to do and @illepic has volunteered to help me get them going. Here's what I see as necessary to cover:

- [ ] Giving Pattern Lab some extra dependencies that Drupal would not get like jQuery, jQuery Once, and domready to name a few.
- [ ] How to make sure that Drupal JS Behaviors integrates in properly.
- [ ] Updating `npm run new` so that when new JS files are created they are set up like we'd like.

### DIFF
--- a/gulpconfig.js
+++ b/gulpconfig.js
@@ -68,6 +68,10 @@ module.exports = {
       ],
     },
   },
+  webpack: {
+    enabled: true,
+    config: require('./webpack.config.js'),
+  },
   patternLab: {
     enabled: true,
     configFile: 'pattern-lab/config/config.yml',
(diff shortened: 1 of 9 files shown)


illepic:
Closing this for now as there is a new webpack branch getting ready to PR against `epic/v10`: https://github.com/phase2/pattern-lab-starter/tree/epic/v10

PR #84, opened 2017-05-06
Add a setup command to facilitate integration with other tooling

tekante:
As pattern lab starter is often used as part of a Drupal theme and a
larger project, a setup command provides the ability for tools like
grunt drupal tasks to request a theme compile which won't fail due to
a missing prior request to install dependencies.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ before_install:
   - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
 install:
   - npm install
-  - npm setup
+  - npm run setup
 before_script:
   - php --version
   - node --version
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ before_install:
   - rm -rf ~/.nvm && git clone https://github.com/creationix/nvm.git ~/.nvm && (cd ~/.nvm && git checkout `git describe --abbrev=0 --tags`) && source ~/.nvm/nvm.sh && nvm install $TRAVIS_NODE_VERSION
 install:
   - npm install
+  - npm setup
 before_script:
   - php --version
   - node --version
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@
 
 ```bash
 npm install # or `yarn install` if you want 3x the speed
-npm setup
+npm run setup
 npm start
 ```
 
--- a/README.md
+++ b/README.md
@@ -12,6 +12,7 @@
 
 ```bash
 npm install # or `yarn install` if you want 3x the speed
+npm setup
 npm start
 ```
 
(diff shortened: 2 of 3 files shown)


tekante:
This pulls the pieces of the diff from https://bitbucket.org/phase2tech/phase2-website/pull-requests/210/introduce-npm-run-build-to-the-theme-and/diff over as a pull request though reading that thread it is unclear rather some adjustments to the postinstall command are desired to remove these items from it.

EvanLovely:
Do we need the `npm run compile` at the end? This will also need to be done (I can't as I'm on my phone): 

- [ ] Remove these setup commands from `postinstall`
- [ ] Add `npm run setup` to readme

tekante:
I've updated the PR but the core question of whether the compile should be removed from the setup command is what is unclear to me from the PR that sparked rolling these upstream. It seems reasonable to me that setup doesn't not imply compile (but does install imply setup?) Since I wasn't actively in the conversation where I think those decisions were made that is where I could use some clarity so I can get this PR to reflect the desired state if it differs from the state of the PR that triggered this submission.

grayside:
From the full project perspective, there are three different operations I am hoping we can cover, and in covering it minimize redundancy.

1. **npm install**: Installs all node dependencies for tools. Since npm install rarely needs to be done and takes a bit of time, I prefer not to see this as part of the build process. Our current practice in the event a repo has two package.json's (typical) is for the root level package.json to run the npm install of the other pakckage.json on postinstall.
2. **build**: Everything else that needs to be done to produce working code. This includes downloading external dependencies and performing necessary transformations/compilations/transpilations on assets and code.
3. **theme helper commands:** This is much less essential than the other two, as I'm not convinced anyone even realizes or needs to know that grunt-drupal-tasks allows running `grunt themes:themename:operation` based on a little Grunconfig.json snippet.

So from this standpoint, I would prefer to see pattern-lab-starter take these practices:

* postinstall: Only clean up files, not automatically run composer install or other commands.
* setup: (Our stand-in for "build") Run composer install, bower install

On the question of compilation, I have no strong feeling whether setup should run compile. We can modify grunt-drupal-tasks to call two different theme commands and existing projects could be setup to run: `npm run setup && npm run compile`.

tekante:
Based on Adam's feedback I've removed the compile from the setup call but I don't think this leaves setup as a corollary for build as it only gets the codebase to a state where working code could be produced at a project level (after the completion of a compilation and a Drupal install). I think that may be a conversation best had in person (and which I think may have already been settled and I just need to catch up on definition of working code). That will leave open the question of what should trigger the compilation at the project level which I'll note in the PR in the relevant project.

EvanLovely:
All this looks good; merging now!

PS Looks like we're really close to being able to pull out that `postinstall` that deletes`*.info` in `node_modules` as well: https://www.drupal.org/node/2329453

tekante:
Awesome, what's the criteria for removing that postinstall hook? The fix you reference looks like it has been included in Drupal since version 8.2.0-beta1 (though any projects that had their settings.php file created before that will need to add in the exclusion patterns).

EvanLovely:
Just feeling like we don't need it for any new projects (as ones already kicked off will still have it). Perhaps we could move it to the docs and say if you get this error, then add this post install script.

PR #93, opened 2017-08-01
Feature/v10/remove bower

madhaze:
Travis CI will fail due to requiring the removal of bundleBower in p2-theme-core v10 https://github.com/phase2/p2-theme-core/tree/epic/v10


anniegreens:
Why are we removing Bower? This is still being used on all projects I have been on.

illepic:
@anniegreens : this is for the v10 release. Projects that are live on pattern-lab-starter v9 will still get the appropriate p2-theme-core (v9).

Follow this chain to the end for the fate of bower: https://github.com/bower/bower/issues/2298. We already have NPM which is a first class source for node tools like Webpack and Gulp, so we're just going from two package management systems to one.

PR #96, opened 2017-08-04
Create license

illepic:


evanmwillhite:
Just noticed this - someone corrected me on [Emulsify](https://github.com/fourkitchens/emulsify/issues/89) for having an MIT license when it appears [anything that extends Drupal needs a GPL2 license](https://www.drupal.org/about/licensing#q7).

illepic:
@evanmwillhite : Interesting. I'm learning about licensing here, but is the basic gist of this that any Drupal theme or module must be GPL2?

grayside:
That's always been my understanding, but MIT is regarded as compatible so maybe you can dual license?

illepic:
Noting here to update in package.json as well. @grayside, I guess we'll just go GPL2 and keep it simple?

PR #107, opened 2017-10-04
Safely joining url and path together

EvanLovely:
Using [`url.resolve()`](https://nodejs.org/api/url.html#url_url_resolve_from_to) is better in case `/` are forgotten in either var.


illepic:
Pulling this work into this PR: https://github.com/phase2/pattern-lab-starter/pull/108. Thanks @EvanLovely !!

PR #122, opened 2017-10-15
Twig namespacing refactor

illepic:
Twig namespacing is now more configurable and works within the gulp.pipe() world better (returns Buffer files).


gulpfile.js
```diff
@@ -27,6 +19,84 @@ const plCompile = require('./tools/tasks/pl-compile')(plPath);
...
+      templates: {
+        root: 'source/_patterns/04-templates',
+        ignore: '/demo',
+      },
+      pages: {
+        root: 'source/_patterns/05-pages',
+        ignore: '/demo',
+      },
+    },
+    // Which files to read and overwrite with namespace info
```
tjheffner: I would switch the order of written files here, or at least include a note that the `'./tools/pattern-lab/config/config.yml'` object is **mandatory** for pattern lab to compile correctly automatically. The .info file only needs to be written for inclusion of patterns in drupal templates

illepic: @tjheffner, just pushed new notes here, let me know what you think.

illepic:
Tagging @EvanLovely here as this is just a refactor and spiritual successor to your twig-namespaces work.

PR #126, opened 2017-10-17
[HOLD][example] integrate vue into webpack + pl

tjheffner:
At this point my current assigned project diverges from immediate needs of PLS v10, but this PR is a mostly-working example of how to include Vue as part of the design system's javascript. 

Pattern Lab components are registered, and can use vue-defined templates in them, see `vue-card/card.vue` & `vue-card/vue-card.twig`

Need to move something around in order to define app at a higher level for multiple widgets on a page, so `<id="app">` or what-have-you only gets used once. Vue doesn't like being mounted to `<html>` or `<body>` anymore. 

The only other "issue" here is that because PL content loads inside an iframe, the vue-devtools chrome debugger isn't properly recognized: 
```
[HMR] Waiting for update signal from WDS...
VM1455 pattern-lab.js:28906 Download the Vue Devtools extension for a better development experience:
https://github.com/vuejs/vue-devtools
VM1455 pattern-lab.js:28916 You are running Vue in development mode.
Make sure to turn on production mode when deploying for production.
See more tips at https://vuejs.org/guide/deployment.html
VM1455 pattern-lab.js:32263 [WDS] Hot Module Replacement enabled.
```
see the last comment here: https://github.com/vuejs/vue-devtools/issues/236 for more info.

This issue has a workaround that we already are familiar with, just open any pl patterns that use vue in a standalone page (without the pattern lab resizing iframe) and the page will detect development-mode vue / the vue devtools debugger pane will appear properly.

In the meantime i'd like to hold this PR open as a working example / reference point to copy integration-related tasks into the docs


illepic:
Keeping this around for posterity.

PR #133, opened 2017-10-18
Make port and host configurable in wpds, make library more dynamic

illepic:


gulpfile.js
```diff
@@ -101,15 +100,14 @@ gulp.task('webpack:watch:scss-to-json', (cb) => {
...
-const localhost = 'http://localhost:8080';
 // Import webpack config for PL
 const wpconfig = require('./webpack.pl.config');
 // Webpack Dev Server config used for local development.
 // See all available config options:
 // https://webpack.js.org/configuration/dev-server/#devserver
 const serverconfig = {
-  publicPath: url.resolve(localhost, wpconfig.output.publicPath), // ie http://localhost:8080/temp
+  host: '0.0.0.0',
+  port: '8080',
```
grayside: This can be overridden from webpack.pl.config?

illepic: There is zero webpack dev server config in our webpack configs. It's either handled at our [webpack-dev-server.js](https://github.com/phase2/pattern-lab-starter/pull/133/files#diff-1763473eec6d663f46bbaecfdf53009e) service or within this gulpfile.js task.

tools/tasks/webpack-dev-server.js
```diff
@@ -12,29 +13,39 @@ let wpds = null; // Hold a reference to Webpack Dev Server when it is created
...
  *  - Added entry points
  */
-module.exports = function startWebpackDevServer(webpackConfig, localhost, devServerConfig) {
+module.exports = function startWebpackDevServer(webpackConfig, devServerConfig) {
   return (cb) => {
     const localWebpackConfig = webpackConfig;
+    const {
+      entry: { 'pattern-lab': plEntry }, // ./source/pattern-lab.js
+      output: { publicPath }, // /temp/
+    } = webpackConfig;
```
thagler: Is this the magic that handles the path issues when served from within a container?

gulpfile.js
grayside: Well, the thing is that binding to 0.0.0.0:80 is really aggressive outside the context of a container.

PR #134, opened 2017-10-19
revamp server watch

illepic:


gulpfile.js
```diff
@@ -130,14 +129,32 @@ const serverconfig = {
...
+  wpds = startWebpackDevServer(wpconfig, serverconfig)(cb);
+});
 
 /**
- * Starts up the Webpack Dev Server with our config from aove
+ * Refresh an active instance of webpack dev server
  */
-gulp.task('webpack:server', webpackdevserver);
+gulp.task('webpack:refresh-server', (cb) => {
+  // Note: wpds lives in scope above this.
```
grayside: What's a wpd?

tools/tasks/webpack-dev-server.js
```diff
@@ -37,18 +36,34 @@ module.exports = function startWebpackDevServer(webpackConfig, devServerConfig)
...
+    return wpds;
   };
+}
+
+/**
+ * Take an instance of webpack dev server, send it random state. Causes reload.
+ * @param server
+ * @param cb
+ */
+function refreshWebpackDevServer(server, cb) {
```
grayside: I've read through this, and from the changeset alone I still don't understand the concept. Maybe it needs more comments, or maybe I'm personally lacking context. Is the idea that that you finish all compilation then send some random querystring at the server to indicate a change instead of using a pure file watch?

anniegreens: @grayside That was the understanding I got from the little walkthrough bloom did this afternoon.

gulpfile.js
illepic: @grayside, note the comment mentioning the parent scope. Var is instantiated on [line 134](https://github.com/phase2/pattern-lab-starter/pull/134/files/f85af2f98fad2537f9854ffab0dd30cd150aafac#diff-b9e12334e9eafd8341a6107dd98510c9R134).

jryanconklin: Shorthand for webpack dev server I believe. Set on 134 and altered.

tools/tasks/webpack-dev-server.js
illepic: Docker won't play nice with us watching compiled file output, nor do we really want to watch that many files just for an indicator of "yep, time to reload the page".

So our change is this:

1. We instantiate a little webpack dev server
1. We return that server somewhere that can hold on to it
1. We can refresh it manually again any time we want

We refresh it manually after the pattern lab command line fully returns successful (rather than watching the pl compiled output directory for files changing). We do this in the gulp task named `webpack:watch:pl-source`.

illepic: And a quick note about the random string: webpack dev server determines whether it needs to do a "hard reload" or "soft reload" based on comparing current state hash to new state hash. We just send over junk and webpack dev server goes "oh snap, hard reload time!". This is only needed for our HTML files generated by PL since they are not on the dependency chain of webpack.

gulpfile.js
illepic: Noting here that with more time and elegance, I'd store the wpds as a singleton. Basically, it would look like:

    // instantiation
    webpackdevserver.start(wpconfig, serverconfig)(cb)

    // Use
    webpackdevserver.reload();

Dammit, @grayside, now I really want to do this.

grayside: I generally avoid acronym variables because reviewing diffs is harder, but I grok now.

tools/tasks/webpack-dev-server.js
grayside: Could we convert some of these comments to inline comments?

illepic:
@grayside : webpack dev server library and implementation is much simpler now.

Issue #114, opened 2017-10-10
Add pa11y accessibility testing to all patterns.


jcar3488:
https://github.com/pa11y/pa11y

Internal notes on potential platform wide approach: https://docs.google.com/document/d/1gXfdl1axMTbe8sNC2z5v3IFn-BiXEIHJaAnVptzkvKg/edit#heading=h.4asc00umu61p

@grayside, what is your opinion about letting us bring tooling into this design system instead of a container approach?

illepic:
PR open here: https://github.com/phase2/pattern-lab-starter/pull/135

grayside:
Well, since you pinged in that way, I'll share medium-sized thoughts about containers, backend tooling, and the questions this approach brings to mind.

I emphasized the significance of a container-based approach because of my general feeling that anything which requires a headless browser seems to involve large file sizes and/or special installation requirements. Purely on the speed of build process, for such cases using a container can be helpful as that piece of the testing can be entirely pre-built.

The reason I aimed for this more from a whole-project standpoint is that we need to have a facility to test accessibility not just for the design system, but also in a rubber-hits-the-editorial-road website.

I want to make sure that all testing gets folded in the CI process of projects that use PLS, currently we emphasize installation, code validation, and "compilation" as steps which allow the theme specific ways to participate in that wholistic/backend perspective of running processes.

From my perspective, here are the questions to dig into a bit:

* Can we find a good way to trigger this kind of testing from grunt-drupal-tasks et al? (I have ideas)
* Can this testing target a Drupal site or just the design system? Can it be flexible enough for local and alternate environments?
* Are there persnickety requirements for pa11y which would make merging this with other node dependencies tricky if it's not tested in that way on an ongoing basis?
* Does it impede the build process performance?

illepic:
Tooling and documentation is in for this.

PR #135, opened 2017-10-22
pa11y accessibility testing against compiled html

illepic:
A new gulp task `test:accessibility` has been created that runs against all appropriate compiled html in dist/public/patterns. It's configured for WCAG2AAA compliance. This task has been added to an overall gulp task called `test`, as well.

Run via `npm run test` or `npm run gulp -- test` and should output something like:

2017-10-22 at 15 12

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # Prerequisites
 
-- [Node](https://nodejs.org) v6 + (v4 might work, no promises)
+- [Node](https://nodejs.org) v6, v8 is required for pa11y testing
 - PHP 5.4, 5.5, 5.6, OR 7 (5.3 might work, no promises)
 - [`composer`](https://getcomposer.org)
 
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -100,7 +100,7 @@ gulp.task('webpack:watch:scss-to-json', (cb) => {
 /**
  * Accessibility test a subset of generated HTML files
  */
-const { pa11yConcurrent } = require('./tools/tasks/gulp-pa11y');
+const { pa11yConsecutive } = require('./tools/tasks/gulp-pa11y');
 
 gulp.task('test:accessibility', () => gulp
   .src([
@@ -108,7 +108,7 @@ gulp.task('test:accessibility', () => gulp
     '!dist/public/patterns/**/*.markup-only.html', // Except the fragment pages
     '!dist/public/patterns/**/index.html', // Except the aggregate, auto-generated pages
   ])
-  .pipe(pa11yConcurrent()));
+  .pipe(pa11yConsecutive()));
 
 /**
  * Webpack config and setup.
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -98,6 +98,19 @@ gulp.task('webpack:watch:scss-to-json', (cb) => {
 });
 
 /**
+ * Accessibility test a subset of generated HTML files
+ */
+const { pa11yConcurrent } = require('./tools/tasks/gulp-pa11y');
+
+gulp.task('test:accessibility', () => gulp
+  .src([
+    'dist/public/patterns/**/*.html', // All html
+    '!dist/public/patterns/**/*.markup-only.html', // Except the fragment pages
+    '!dist/public/patterns/**/index.html', // Except the aggregate, auto-generated pages
+  ])
+  .pipe(pa11yConcurrent()));
+
+/**
  * Webpack config and setup.
  */
 // Import webpack config for PL
@@ -168,6 +181,13 @@ gulp.task('compile', gulp.series([
   'compile:scss-to-json',
   'compile:twig-namespaces',
   'compile:pl',
+]));
+
+/**
+ * Run tests against non-webpack assets
+ */
+gulp.task('test', gulp.parallel([
+  'test:accessibility',
 ]));
 
 /**
--- a/tools/tasks/gulp-pa11y.js
+++ b/tools/tasks/gulp-pa11y.js
@@ -23,7 +23,7 @@ const pa11yConfig = _.assign(pallyJson, logConfig);
  * Run pa11y testing one after the next
  * @returns {*}
  */
-function pa11yConcurrent() {
+function pa11yConsecutive() {
   function bufferContents(vinylFile, encoding, callback) {
     // ignore empty files
     if (vinylFile.isNull()) {
@@ -87,6 +87,6 @@ function pa11yParallel() {
 }
 
 module.exports = {
-  pa11yConcurrent,
+  pa11yConsecutive,
   pa11yParallel,
 };
--- a/tools/tasks/gulp-pa11y.js
+++ b/tools/tasks/gulp-pa11y.js
@@ -6,17 +6,18 @@ const reporter = require('pa11y-reporter-cli'); // As pa11y 5 stabilizes, we can
 const _ = require('lodash');
 const through = require('through2');
 
-// Grab project root pa11y.json as object, decorate with some boilerplate pa11y logging
-const pa11yConfig = _.assign(
-  JSON.parse(fs.readFileSync(path.join(process.cwd(), 'pa11y.json'))),
-  {
-    log: {
-      debug: console.log,
-      error: console.error,
-      info: console.log,
-    },
+// Grab project root pa11y.json as object
+const pallyJson = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'pa11y.json')));
+// Custom logging tools for pa11y
+const logConfig = {
+  log: {
+    debug: console.log,
+    error: console.error,
+    info: console.log,
   },
-);
+};
+// Smash them together
+const pa11yConfig = _.assign(pallyJson, logConfig);
 
 /**
  * Run pa11y testing one after the next
--- a/tools/tasks/gulp-pa11y.js
+++ b/tools/tasks/gulp-pa11y.js
@@ -1,0 +1,91 @@
+const fs = require('fs');
+const path = require('path');
+
+const pa11y = require('pa11y');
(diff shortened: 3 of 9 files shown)


illepic:
Noting here: pa11y v5 testing requires Node 8.

README.md
```diff
@@ -2,7 +2,7 @@
 
 # Prerequisites
 
-- [Node](https://nodejs.org) v6 + (v4 might work, no promises)
+- [Node](https://nodejs.org) v6, v8 is required for pa11y testing
```
grayside: With pa11y added as a dependency, will it even install on Node 6, or is the Node 8 dependency specifically in trying to execute pa11y?

gulpfile.js
```diff
@@ -98,6 +98,19 @@ gulp.task('webpack:watch:scss-to-json', (cb) => {
...
 /**
+ * Accessibility test a subset of generated HTML files
+ */
+const { pa11yConcurrent } = require('./tools/tasks/gulp-pa11y');
+
+gulp.task('test:accessibility', () => gulp
+  .src([
+    'dist/public/patterns/**/*.html', // All html
+    '!dist/public/patterns/**/*.markup-only.html', // Except the fragment pages
+    '!dist/public/patterns/**/index.html', // Except the aggregate, auto-generated pages
```
grayside: Per my comment on #114, I suppose the answer to "how can we test against Drupal", would be an alternate execution approach that queries URLs instead of files?

tools/tasks/gulp-pa11y.js
```diff
@@ -0,0 +1,92 @@
+const fs = require('fs');
```
grayside: Should this file be considered for a standalone library?

tools/tasks/gulp-pa11y.js
```diff
@@ -0,0 +1,92 @@
...
+const path = require('path');
+
+const pa11y = require('pa11y');
+const reporter = require('pa11y-reporter-cli'); // As pa11y 5 stabilizes, we can pull this off pa11y
+const _ = require('lodash');
+const through = require('through2');
+
+// Grab project root pa11y.json as object
+const pallyJson = JSON.parse(fs.readFileSync(path.join(process.cwd(), 'pa11y.json')));
+// Custom logging tools for pa11y
```
grayside: Why do we need custom logging tools? Shouldn't CLI logging be part of why we are using the CLI reporter?

tools/tasks/gulp-pa11y.js
```diff
@@ -0,0 +1,92 @@
...
+/**
+ * Run pa11y testing one after the next
+ * @returns {*}
+ */
+function pa11yConcurrent() {
+  function bufferContents(vinylFile, encoding, callback) {
+    // ignore empty files
+    if (vinylFile.isNull()) {
+      callback();
+      return;
```
grayside: That totally seems reasonable, does something else report in on the obvious build problem of files showing up empty?

tools/tasks/gulp-pa11y.js
```diff
@@ -0,0 +1,92 @@
...
+  },
+};
+// Smash them together
+const pa11yConfig = _.assign(pallyJson, logConfig);
+
+/**
+ * Run pa11y testing one after the next
+ * @returns {*}
+ */
+function pa11yConcurrent() {
```
grayside: Don't you mean pa11ySequential or in node terminology `pa11ySync` short for pa11ySynchronous?

README.md
illepic: It blows up on install on Node v6. Two options going forward:

1. Lock hard requirement to Node v8.
1. Babel transpile all tooling js

gulpfile.js
illepic: Indeed, the goal for this PR is *just* PL-rendered HTML. But the tools are in and `gulpfile.js` is ready for new entries to test a local/remote Drupal site.

tools/tasks/gulp-pa11y.js
illepic: I need clarification here.

tools/tasks/gulp-pa11y.js
illepic: BLARG, I meant "consecutive".

tools/tasks/gulp-pa11y.js
illepic: This is boilerplate pa11y and I'd definitely like to look into cleaning this up in the future. For instance, you'll notice that we do this to output log info:

```shell
console.log(reporter.results(result));
```

With more research, I'm guessing it's actually more like:

```shell
pa11y.log.info(reporter.results(result));
```

tools/tasks/gulp-pa11y.js
illepic: Nope, this is just gulp's `gulp.src()` mechanism. If a file exists, it gets sent in regardless. I'm sure there's some YAGP (Yet Another Gulp Plugin) that could let use do this instead:

```js
gulp.task('test:accessibility', () => gulp
  .src([
    'dist/public/patterns/**/*.html',
    ...
  ])
  .pipe(ignoreEmpty()) // This plugin probably exists
  .pipe(pa11yConcurrent()));
```

README.md
grayside: See #136, transpiling all our dependencies sounds insane. I say we either make the addition of pa11y an add-on (Run the script/generator/etc to add pa11y, which we can do as a step of a travis run), or double-down on Node 8 being the LTS release of new projects.

gulpfile.js
grayside: Okay, I started to write up a whole thing, but if this is at least open to it, we have the basis to discuss as a follow-up. I'm not convinced this is the right place for Drupal testing, but as long as it's a possible place we can work it out in The Future.

tools/tasks/gulp-pa11y.js
grayside: E.g., github.com/phase2/gulp-pa11y

tools/tasks/gulp-pa11y.js
grayside: ✅

tools/tasks/gulp-pa11y.js
grayside: Part of my point is that since empty files at the end of a pipeline are obviously an error of some kind, shouldn't we flag those with a warning? Or does gulp do that already?

tools/tasks/gulp-pa11y.js
illepic: Ah, we're talking about the whole file (thought we were talking about `fs`).

So, bringing gulp-pa11y.js, gulp-sass2json.js, and gulp-twig-namespaces.js to some kind of individual package would be great, BUT these things absolutely need to live as close to this project as possible.

Monorepo time? Lerna? For the time being, I'm OK with these just hanging out in `tools/tasks` since I consider the Pattern Lab Starter as much a node app as a Drupal theme.

tools/tasks/gulp-pa11y.js
illepic: Updated.

tools/tasks/gulp-pa11y.js
grayside: I don't see this as a monorepo thing, I just see this as being clearly not all that tied to PL and possibly of use to other projects, so shouldn't it be it's own Open Source project with an issue queue that PLS depends on? Admittedly, I don't see whatever needs it to be kept close to PLS and it's not a blocking point.

tools/tasks/gulp-pa11y.js
illepic: I don't really think empty files at the end of a pipeline is technically an error. That `isNull()` check is being provided every single file in a glob so it's rational to assume some files will be empty and to just skip them.

tools/tasks/gulp-pa11y.js
illepic: One of the big goals of moving to Webpack and folding in p2-theme-core into PLSv10 was to keep all the tooling as close at hand as possible. And, honestly, it would more likely need to be a PR against a project like this instead: https://github.com/dreamzmaster/gulp-pa11y

Let's give it some time out in the world to see how it improves.

README.md
illepic: We can babel-compile just our node libraries (not the entire node_modules).

grayside: Is this a Node 8 requirement of our node libraries or of pa11y 5? I'm confused.

illepic: @grayside, pa11y v5 uses async/await. This feature is not available in Node <8. Our `gulp-pa11y.js` script uses pa11y v5 and thus is locked to node v8.

gulpfile.js
```diff
@@ -171,6 +184,13 @@ gulp.task('compile', gulp.series([
 ]));
 
 /**
+ * Run tests against non-webpack assets
+ */
+gulp.task('test', gulp.parallel([
+  'test:accessibility',
```
tjheffner: our test task should probably run the full lint too (ie including config files). unless testing is different from validation in this case

tools/tasks/gulp-pa11y.js
tjheffner: just seconding that i don't think it's out of place here, doubly so considering all of the "accessibility best practice" default patterns we want to fold in, this seems like a logical extension to include in the pls toolkit

gulpfile.js
illepic: A couple of things here:

There's a hard line between what Gulp does and what WebPack does. Basically, Gulp handles all assets that we can't get into WebPack.

1. We do linting on asset code via webpack, and have an npm script for everything else
1. This accessibility testing touches non-webpack assets, thus gets a Gulp task. Though there is an npm script for single file testing

Basically, I don't want to build out a gulp task for linting we already have done via WebPack.

illepic:
I'm merging, as there's no more outstanding code fixes. Node v8 locally, yall.

PR #140, opened 2017-10-24
Feature/v10/add bootstrap

rloos289:


source/_data/data.json
```diff
@@ -6,6 +6,7 @@
   "bodyClass": "pl",
   "title": "Faker.words(2, true)",
   "text": "Faker.words(3, true)",
+  "cardimage" : "Faker.imageURL(318, 180, 'abstract', true, 'Some Text', true)",
```
illepic: Let's take a look at using the holder.js library for this. You can see an implementation for this in the `_img.twig` file that needs some cleanup.

source/_patterns/01-atoms/button/_button.twig
```diff
@@ -0,0 +1,32 @@
...
+  See https://getbootstrap.com/docs/4.0/components/buttons/ for more details
+#}
+
+<{{ button_element|default('button') }}
+type="{{ button_type }}"
+class="btn btn-{% if button_outline == TRUE  %}outline-{% endif %}{{ button_color|default('primary') }}
+  {% if button_size %} btn-{{ button_size }}{% endif %}
+  {% if button_block %} btn-block{% endif %}
+  {% if button_active %} active {% endif %}"
+{% if button_active == TRUE  %} aria-pressed="true" {% endif %}
```
illepic: Odd switch to explicit `TRUE` here.

source/_patterns/01-atoms/button/_button.twig
```diff
@@ -0,0 +1,32 @@
...
+  button_disabled: Boolean: True|False set button to appear disabled
+  button_text: String: Display text on the button
+
+  See https://getbootstrap.com/docs/4.0/components/buttons/ for more details
+#}
+
+<{{ button_element|default('button') }}
+type="{{ button_type }}"
+class="btn btn-{% if button_outline == TRUE  %}outline-{% endif %}{{ button_color|default('primary') }}
+  {% if button_size %} btn-{{ button_size }}{% endif %}
```
illepic: I'd like you to take a look at recent work on projects for how we can load up these attributes as a concatenated string in a `{% set %}` block. This vastly simplifies this markup.

source/_patterns/01-atoms/button/_button.twig
```diff
@@ -0,0 +1,32 @@
...
+  {% if button_block %} btn-block{% endif %}
+  {% if button_active %} active {% endif %}"
+{% if button_active == TRUE  %} aria-pressed="true" {% endif %}
+  {% if button_toggle == TRUE  %} data-toggle="button" {% endif %}
+  {% if button_link == TRUE  %} href="{{ button_link }}" {% endif %}
+  {% if button_value == TRUE  %} value="{{ button_value }}" {% endif %}
+  {% if button_disabled == TRUE  %} disabled {% endif %}>
+{{ button_text }}
+</{{ button_element|default('button') }}>
```
illepic: Use `set` so we're not double `|default()`'ing ("don't repeat yourself)

source/_patterns/01-atoms/button/demo/buttons.twig
```diff
@@ -0,0 +1,50 @@
+<h1>Button colors</h1>
+{% include '@atoms/button/_button.twig' with button_primary %}
```
illepic: Let's use a loop in twig instead of all this manually.

source/_patterns/01-atoms/button/index.js
```diff
@@ -0,0 +1,12 @@
+// import $ from 'jquery';
+// might be good to force in some jquery here
+import button from 'bootstrap/js/src/button';
```
illepic: Verifying: this works?

source/_patterns/01-atoms/button/index.js
```diff
@@ -0,0 +1,12 @@
...
+
+require('./button.scss');
+
+export default {
+  button: [button],
```
illepic: What is this key?

source/_patterns/02-molecules/card/_card.scss
```diff
@@ -1,5 +1,9 @@
 @import '../../00-base/colors';
```
illepic: I'd like to see _card.scss require button.scss explicitly, if the card uses a button.

source/_patterns/01-atoms/button/index.js
```diff
@@ -0,0 +1,12 @@
...
+  name: 'button_primary',
```
illepic: Is this actually `button_primary`? Or just `button`?

source/_patterns/02-molecules/card/card.js
```diff
@@ -1,14 +1,10 @@
-import $ from 'jquery';
```
illepic: `card.js` should bring over `button.js`, and any other components it needs.

source/design-system.js
```diff
@@ -1,13 +1,14 @@
 import card from './_patterns/02-molecules/card/card';
 import homepage from './_patterns/05-pages/homepage';
-
+import button from './_patterns/01-atoms/button';
```
illepic: Let's import in order. Move this to top.

source/scss/_config.scss
```diff
@@ -21,8 +21,20 @@
 //  xxl: $bp--xxl
 //) !default;
 
+//bootstrap setup
+@import "~bootstrap/scss/functions";
+@import "~bootstrap/scss/variables";
+@import "~bootstrap/scss/mixins";
+@import "~bootstrap/scss/print";
```
illepic: Hard rule: `_config.scss` produces **no printing css**.  So we'll cut it above `print`. This means that reboot, type, etc, have to be pulled in elsewhere in our `00-base` folder, or in `design-system.js`.

webpack.shared.config.js
```diff
@@ -82,6 +82,12 @@ module.exports = {
       filename: '[name].styles.css',
       allChunks: true,
     }),
+    new webpack.ProvidePlugin({
+      $: 'jquery',
+      jQuery: 'jquery',
+      'window.jQuery': 'jquery',
+      Popper: ['popper.js', 'default'],
```
illepic: Comment here on what this does.

source/_patterns/01-atoms/button/_button.twig
illepic: (noting that this won't matter when we move to a cleaner `set` syntax.

source/_patterns/01-atoms/button/index.js
rloos289: Yep, this works. The main thing I was looking for here was that "toggled" buttons stay toggled (http://getbootstrap.com/docs/4.0/components/buttons/#toggle-states)

source/_patterns/01-atoms/button/index.js
rloos289: This is to pull in button.js. If I don't include the brackets then the linter hits me with this error:
`8:3  error  Expected property shorthand  object-shorthand`

illepic:
Noting that a big issue we're trying to solve here is duplication of css from modular imports. In other words:

`_button.scss` produces some css. And `_card.scss` needs to ensure it has `_button.scss`. Since we import **all** components in `design-system.js`, we end up with duplicated button css in the rendered output.

Looking into PostCSS plugins to get around this.

illepic:
Progress notes:

We have deduping capabilities after Sass renders (not optimal, but gets the job done) by adding this key to the css-loader:

```js
minimize: {
  discardDuplicates: true,
}
```

HOWEVER, this does now work when importing scss individually in js files. Essentially, webpack is not sharing **context** between scss imported. 

So we get deduping per scss file (and scss imported in scss). But we don't get deduping across sass files imported in js.

For now, looking to solve this.

illepic:
Progress report:

Importing dependent components via **JavaScript** allows our sass dependencies to behave as we expect with no dupes.

source/_patterns/00-base/_mixins.scss
```diff
@@ -1,3 +1,17 @@
+/// Mixins
+
+/// Export Sass Data to JSON in `path/to/export/folder/` folder
+/// @param {String} $filename - ie `mystuff.json`
+/// @param $var - What to turn into JSON
+/// @example scss
+///   @include export-data-to-lib('filename.json', $sass-map);
+@mixin export-data($filename, $var, $options: ()) {
```
tjheffner: these are good comments, but please please please get a section in the docs on this + where it's used + why in detail. i guarantee it will make people mad/confused otherwise

webpack.shared.config.js
```diff
@@ -5,7 +5,10 @@ const StyleLintPlugin = require('stylelint-webpack-plugin');
 const ExtractTextPlugin = require('extract-text-webpack-plugin');
 const IconfontWebpackPlugin = require('iconfont-webpack-plugin');
 const autoprefixer = require('autoprefixer');
-
+const sassExportData = require('@theme-tools/sass-export-data')({
```
tjheffner: where does @theme-tools come from?

source/_patterns/00-base/_mixins.scss
illepic: This is a little shifty as a feature right now (I'm not certain it will survive in its current state), so I'm waiting to spend the time documenting until we get this documenting fonts, typog, grids, etc.

webpack.shared.config.js
illepic: `package.json`

source/_patterns/02-molecules/card/_card.twig
```diff
@@ -0,0 +1,109 @@
...
+  see https://getbootstrap.com/docs/4.0/components/card/ for more details
+#}
+
+{% set classes = [
+  card_width ? 'w-' ~ card_width,
+  text_alignment ? 'text-' ~ text_alignment,
+  text_color ? 'text-' ~ text_color,
+  card_background ? 'bg-' ~ card_background,
+  card_border ? 'border-' ~ card_border,
+  'mb-3 '
```
illepic: Definitely don't want `mb-3` here.

source/_patterns/01-atoms/grid/demo/grid-demo.twig
```diff
@@ -0,0 +1,47 @@
+<div class="container">
+  <h1>One column</h1>
```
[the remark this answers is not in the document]
illepic: Nope, doesn't matter. Nice. Yay bootstrap.

source/_patterns/01-atoms/button/index.js
```diff
@@ -0,0 +1,19 @@
...
+// Import our custom sass, includes Bootstrap sass
+import './_button.scss';
+
+// Custom logic for buttons here
+
+export default {
+  name: 'button',
+  deps: [base, Button],
+  enable() {
+    console.log('Button js ran!');
```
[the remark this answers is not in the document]
rloos289: Yep, the button.js functionality works when its imported this way.

PR #141, opened 2017-10-25
Travis now only tests on Node v8

illepic:


illepic:
@geoff-winner still need the `engines` key in package.json, this PR doesn't negate that open issue.

PR #144, opened 2017-10-30
Feature/v10/jumbotron

jryanconklin:
This PR will:

- Update `.gitignore` to deal with unwanted system files
- Update package-lock
- Add modularized Bootstrap 4 jumbotron component

The jumbotron-fluid implementation requires an inner `container` class wrapping the content. To account for that here I have two if statements in Twig for `jumbotron_fluid` but can imagine there's a better way to accomplish this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,8 @@ source/_data/scss*.json
 # OS generated files
 .DS_Store
 .DS_Store?
+._.DS_Store
+._.DS_Store?
 
 # Dependencies
 bower_components
@@ -15,6 +17,7 @@ node_modules
 # Generated files
 *.sass-cache/
 .eslintcache
+*unison.*
 
 #IDE plugins
 .idea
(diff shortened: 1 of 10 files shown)


.gitignore
```diff
@@ -15,6 +17,7 @@ node_modules
 # Generated files
 *.sass-cache/
 .eslintcache
+*unison.*
```
illepic: 👏

source/_patterns/02-molecules/jumbotron/index.js
```diff
@@ -0,0 +1,24 @@
...
+import image from 'atoms/image';
+import button from 'atoms/button';
+
+// Module styles
+import './_jumbotron.scss';
+
+export default {
+  name: 'jumbotron',
+  deps: [base, image, button],
+  enable($context, { enableHolder }) {
```
illepic: @jryanconklin remove holder destructuring and logic below. This is only needed in the image component.

Issue #137, opened 2017-10-23
Add Faker Twig Function


jryanconklin:
See Issue on [php twig edition](https://github.com/pattern-lab/edition-php-twig-standard/issues/7) for bones of Twig function using Faker. Adapting this would use of Faker within Twig as:

```
{{ faker('property', 'options' }}
```

This is likely impacted by the issues currently in [`pattern-lab/plugin-php-faker`](https://github.com/pattern-lab/plugin-php-faker/issues). For example, when working with this as `{{ faker('ImageUrl', 200, 200)  }}`, only the first argument is honored, returning: `https://lorempixel.com/200/480/?57256`. Note there is an open PR for plugin [issue #4](https://github.com/pattern-lab/plugin-php-faker/issues/4)

illepic:
@jryanconklin do we want to leave this open?

jryanconklin:
No, let’s go ahead and close it.

PR #145, opened 2017-10-30
Issue #137: Add Faker Twig Function

jryanconklin:
Twig function for simple Faker use in templates, from comments on [php twig edition](https://github.com/pattern-lab/edition-php-twig-standard/issues/7). Allows for:

```
{{ faker('property', 'options' }}
```

This comes with some caveats of use, but think this would still be a good addition. See issue #137 for original write-up.

### DIFF
--- a/source/_twig-components/functions/faker.function.php
+++ b/source/_twig-components/functions/faker.function.php
@@ -1,0 +1,10 @@
+<?php
+
+$function = new Twig_SimpleFunction('faker', function ($string, $options="") {
+  $faker = Faker\Factory::create();
+  if ($options != "") {
+    return $faker->$string($options);
+  } else {
+    return $faker->$string;
+  }
+});


source/_twig-components/functions/faker.function.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+$function = new Twig_SimpleFunction('faker', function ($string, $options="") {
+  $faker = Faker\Factory::create();
+  if ($options != "") {
+    return $faker->$string($options);
```
grayside: Relying on options to indicate whether $string is a callable is dicey. Is there a reason to not instead or additionally use http://php.net/is_callable?

source/_twig-components/functions/faker.function.php
```diff
@@ -0,0 +1,10 @@
...
+  } else {
```
grayside: The else here seems unnecessary given the above is a return statement. So...

```
if ($thing) {
  return "magic";
}

return "science";
```

source/_twig-components/functions/faker.function.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+$function = new Twig_SimpleFunction('faker', function ($string, $options="") {
+  $faker = Faker\Factory::create();
+  if ($options != "") {
```
grayside: Why specifically check for an empty string? `!empty()` seems like it would cover more user error cases.

source/_twig-components/functions/faker.function.php
```diff
@@ -0,0 +1,10 @@
+<?php
+
+$function = new Twig_SimpleFunction('faker', function ($string, $options="") {
```
illepic: In the comments here, hit me with demo implementation in DocBlock format.

source/_twig-components/functions/faker.function.php
illepic: We have a default value setting `$options` to empty string. Can't all this just be simplified down to:

`return $faker->$string($options)`

No switch, no check?

jryanconklin: Probably, I wasn't sure how much effort I should put into this since I was copy/pasting this from an issue queue. Plus the issues with faker generally. Thought of the PR more as a, "this is a route we could go if we can get the Faker plugin fixed."

jryanconklin:
Here's another path we can take with this that may be a bit better:
https://packagist.org/packages/alanablett/twig-faker

I believe we'd need to add this via Composer to `tools/pattern-lab`

illepic:
Going to close this one down for the time being, but we might revisit better faker approaches in the future.

PR #147, opened 2017-11-01
Feature/v10/breadcrumbs

jryanconklin:
This PR adds the Breadcrumb component from Bootstrap, see:
https://getbootstrap.com/docs/4.0/components/breadcrumb/

There's an TODO on this for dealing with active menu items / aria-current as that might need more work once the Drupal env is ready. Bootstrap uses `active` for this whereas `is-active` is what I'm guessing we'd get from Drupal.


source/_patterns/02-molecules/breadcrumb/_breadcrumb.twig
```diff
@@ -0,0 +1,23 @@
...
+  @TODO: Add ability to toggle 'active' class and aria-current attribute to page.
+
+  see https://www.w3.org/TR/wai-aria-practices/#breadcrumb
+
+  see https://getbootstrap.com/docs/4.0/components/breadcrumb/ for more details
+#}
+
+<nav aria-label="breadcrumb" role="navigation">
+  <ol class="breadcrumb">
+    {% if breadcrumb_items %}
```
illepic: I think it's safe to remove this check. It won't iterate if nonexistent.

source/_patterns/02-molecules/breadcrumb/_breadcrumb.scss
```diff
@@ -0,0 +1,18 @@
+// All component Sass needs non-printing base config
```
illepic: Let's make this an atom. Even though it renders other things, it doesn't necessarily include any other Patterns.

source/_patterns/02-molecules/breadcrumb/index.js
```diff
@@ -0,0 +1,21 @@
+/**
+ * Breadcrumb
+ */
+
+// Module dependencies
+import base from 'base';
+import $ from 'jquery';
```
illepic: Probably don't need jQuery here. Also, it should go above local imports.

jryanconklin:
@illepic Updated!

PR #148, opened 2017-11-01
Feature/v10/layout_examples

jryanconklin:
This PR adds:
* A Demo of Containers Layout under atoms/grid.
* Give basic usage info in Grid Demo. Mainly linking to Bootstrap docs due to the level of detail.


source/_patterns/01-atoms/grid/demo/layout-demo.twig
```diff
@@ -0,0 +1,17 @@
+<h1>Layout Demo</h1>
+<br>
+<div class="container bg-secondary" style="height: 20rem;">
+  <h1 class="text-white">Container Standard</h1>
+  <p>
+    Containers are the most basic layout element in Bootstrap and are required when using our default grid system.
+    Choose from a responsive, fixed-width container (meaning its <code>max-width</code> changes at each breakpoint) or fluid-width (meaning it’s <code>100%</code> wide all the time).
+    While containers can be nested, most layouts do not require a nested container.
```
illepic: I don't think we can have containers within containers, or at least it is highly recommended against.

source/_patterns/01-atoms/grid/demo/layout-demo.twig
```diff
@@ -0,0 +1,17 @@
+<h1>Layout Demo</h1>
```
illepic: Put this in a .container to it doesn't touch left viewport.

jryanconklin:
@illepic Updated! The note on nested containers is directly from Bootstrap. I added a caveat stating that it's possible but not recommended.

PR #152, opened 2017-11-05
ICONS

illepic:
This PR does some big stuff.

1. First: **The local server hosts from http://0.0.0.0:8080/ and to access PL you visit http://0.0.0.0:8080/pl/**. This is documented in the terminal output, but you can miss this if you're not looking for it.
1. Compiles SVGs into a font and base64-inlines the output into CSS. All prior icon mixins and classes work the same way. See the demo for details
1. Brings over a favicon autocompiler
1. Includes a working example of how to bundle and move static assets from source to dist. Check out `@atoms/image/demo/logo-img.md` for details.
1. Includes an example of how to inline svg


tools/pattern-lab/config/config.yml
```diff
@@ -18,7 +19,7 @@ patternStates:
 styleGuideExcludes: ''
 cacheBusterOn: 'true'
 exportDir: export
-publicDir: ../../dist/public/
+publicDir: ../../dist/pl/
```
jryanconklin: This is a change we'll probably want to make sure is included in the docs.

source/styleguide/_scss2json.scss
```diff
@@ -6,18 +6,18 @@
 
 // Customized Sass Mapping
 // Pulled from Bootstrap's _variables.scss
-$font-families: (
+$font-families-demo: (
```
jryanconklin: Nice, this is clearer.

source/pattern-lab.js
```diff
@@ -25,7 +25,7 @@ const settings = {
 };
 
 // Let's just execute everything and pass in $(document), settings
-_.forEach(designSystem, (component, name) => {
+_.forEach(ds, (component, name) => {
```
jryanconklin: Agree shorthand might be better here, but are we sure `ds` is descriptive enough to understand for new people looking at this file?

source/styleguide/_scss2json.scss
```diff
@@ -6,18 +6,18 @@
 
 // Customized Sass Mapping
```
jryanconklin: We should document this file more than I had initially via comments.

gulpfile.js
```diff
@@ -120,7 +120,7 @@ const wpconfig = require('./webpack.pl.config');
 const serverconfig = {
   host: '0.0.0.0',
   port: '8080',
-  contentBase: path.resolve(__dirname, 'dist/', 'public/'), // ie dist/public
+  contentBase: path.resolve(__dirname, 'dist/'), // ie dist/pl
```
grayside: `// ie dist/pl` means nothing to me. Would a reader need more context? SOCRITICALNOTREALLY

package.json
```diff
@@ -77,8 +77,10 @@
     "stylelint-scss": "^1.4.3",
     "stylelint-webpack-plugin": "^0.9.0",
     "through2": "^2.0.3",
+    "url-loader": "^0.6.2",
     "webpack": "^3.8.1",
     "webpack-dev-server": "^2.9.4",
+    "webpack-iconfont-plugin": "0.0.2",
```
grayside: Per some other comment somewhere, do these new icon processing libraries have a dependency on the fontconfig server package or other packages?

package.json
```diff
@@ -61,12 +61,12 @@
...
     "eslint-plugin-react": "^7.4.0",
     "extract-text-webpack-plugin": "^3.0.2",
+    "favicons-webpack-plugin": "0.0.7",
     "file-loader": "^1.1.5",
     "glob": "^7.1.2",
     "gulp": "github:gulpjs/gulp#4.0",
-    "iconfont-webpack-plugin": "^1.1.0",
     "js-yaml": "^3.10.0",
-    "node-sass": "^4.5.3",
+    "node-sass": "^4.6.0",
```
grayside: Minor version bump! Does that mean there are new goodies to be leveraged?

patternlab.info.yml
```diff
@@ -21,7 +21,7 @@ component-libraries:
     paths:
       - source/_patterns/01-atoms
       - source/_patterns/01-atoms/grid
-      - source/_patterns/01-atoms/icons
+      - source/_patterns/01-atoms/icon
```
grayside: Sweet pluralization consistency!

source/_data/data.json
```diff
@@ -2,7 +2,7 @@
   "pl": true,
   "BUILD_TARGET": "pl",
   "paths": {
-    "pl_theme_root": "../../../.."
+    "assets": "../../../assets"
```
grayside: So if you wanted to move the build target of the assets, you'd edit this here?

Reasonable to consider switching this to a yaml file?

source/_meta/_00-head.twig
```diff
@@ -5,21 +5,14 @@
 		<meta charset="UTF-8">
 		<meta name="viewport" content="width=device-width" />
 
+		<link rel="icon" type="image/png" href="{{ paths.assets }}/favicons/favicon-16x16.png" sizes="16x16">
+		<link rel="icon" type="image/png" href="{{ paths.assets }}/favicons/favicon-32x32.png" sizes="32x32">
```
grayside: Per stuff like https://realfavicongenerator.net/, only two icons here seems light. Should we cover more stuff as a baseline? Nice icons on mobile devices seem like something we should have by default.

source/pattern-lab.js
grayside: Potentially painful for blind refactoring. thunderSnow?

tools/tasks/webpack-dev-server.js
```diff
@@ -44,7 +44,7 @@ function startWebpackDevServer(webpackConfig, devServerConfig) {
         cb(err);
       }
 
-      console.info(`Listening at ${localHost}`);
+      console.info(`Listening at ${localHost}. PL located at ${localHost}/pl`);
```
grayside: Not that we should pull it, but keep in mind that this will not be correct when run via docker container. As a future enhancement, being able to override this message would be a nice touch.

package.json
anniegreens: LOL, herein lies the issue with names:

- removes `"iconfont-webpack-plugin"`
- adds `"webpack-iconfont-plugin"`

source/_meta/_00-head.twig
anniegreens: Is there a reason we don't use `.ico` with multiple sizes included within? I suppose the HTML5 way is preferred here. 

This may be slightly outdated, but perhaps we should add some mobile/touch ones: https://css-tricks.com/favicon-quiz/

source/_data/data.json
tjheffner: Yes, this should get updated if/whenever the build target of the assets changes. But this is a global pattern lab helper, the actual build target is set inside webpack.shared.config.js (the key titled "output"). I imagine drupal-theme.js or the .theme file will set `path.assets` to `\Drupal::theme()->getActiveTheme()->getPath();`


To the second point, I would like to see all of source/_data converted to .yml eventually, although I don't know if that's an option in the case of `scssVariables.json` and how that's being generated. Right now our practices aren't consistent in either direction.

gulpfile.js
tjheffner: we'll be sure to get more notes around this in the docs repo, but this should hopefully be enough in context from the webpack doc link in the comments above and the terminal output: `Listening at http://0.0.0.0:8080. PL located at http://0.0.0.0:8080/pl` what contentBase is / where pl is coming from.

source/_data/data.json
illepic: There is a ticket to convert all human-created _data files to yaml. Robot-created _data should stay json. More than a few _data files should be deleted or consolidated.

source/_meta/_00-head.twig
illepic: Note: the webpack plugin has the ability to do ALL. This will be captured in FE docs eventually, but for now it's the two most common formats.

source/_patterns/01-atoms/grid/_grid-1-up.twig
```diff
@@ -3,13 +3,10 @@
   in a .container div. We leave this up to the implementor manually for flexibility.
 #}
 
-{# For example, .container would go here:#}
-{#<div class="container">#}
-
+{% if container %}<div class="container">{% endif %}
```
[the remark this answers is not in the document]
illepic: @jryanconklin, mind spinning up a ticket for this guy? We can work out the 3 scenarios:

1. No `.container`
1. `.container`
1. `.container-fluid`

source/pattern-lab.js
illepic: `ds` is explicitly aliased from the `design-system.js` component. However, I'll be looking into a better way of exporting all patterns fro design-system, since it currently exports a `designSystem` object causing linter warnings if we attempt to alias locally as the same name.

source/_patterns/01-atoms/grid/_grid-1-up.twig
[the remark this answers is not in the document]
jryanconklin: Ticket created in Jira, PLP-26

source/_meta/_00-head.twig
illepic: Roger, but at this point it fills the webpack bundle output in terminal and is **really** distracting. I'll make a ticket to capture favicons approach.

source/_patterns/01-atoms/grid/_grid-1-up.twig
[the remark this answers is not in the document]
illepic: Ah, I've been putting these in the github issue queue to keep work close to code. Update that jira ticket with the github issue?

tools/pattern-lab/config/config.yml
illepic: Indeed, but holding off as we experiment and stabilize.

tools/tasks/webpack-dev-server.js
illepic: Captured here: https://github.com/phase2/pattern-lab-starter/issues/158. Should just be able to capture an environment var.

source/pattern-lab.js
illepic: Captured here: I've got an idea for a cleaner, less object-y approach to components: 

https://github.com/phase2/pattern-lab-starter/issues/160

PR #166, opened 2017-11-09
JS refactor

illepic:
- All js is now much more es6 module-y
- gulp sass2json is removed, but saved here: https://gist.github.com/illepic/b4c87b988b232b8ad65e54c1ed51f1fe
- docs folder removed, but added to `triage` folder here: https://github.com/phase2/frontend-docs/tree/master/docs/triage
- drupal and pattern lab entry points have been moved to the root clear and consistent folders: `app-drupal` and `app-pl`. All tool has been updated to reflect this.


README.md
```diff
@@ -8,7 +8,7 @@
 
 # QuickStart
 
-[Download this repo](https://github.com/phase2/pattern-lab-starter/archive/master.zip), rename folder to `patternlab` and place in themes directory if integrating with Drupal 8, and then `cd` into folder before running below commands. To name it something else, see [docs/setup.md](docs/setup.md).
+[Download this repo](https://github.com/phase2/pattern-lab-starter/archive/master.zip), rename folder to `patternlab` and place in themes directory if integrating with Drupal 8, and then `cd` into folder before running below commands.
```
jryanconklin: Interesting, I hadn't thought about this yet. This solves my q from yesterday about adding theme hooks. There's no naming morass because we're not renaming. Very cool.

source/_patterns/01-atoms/breadcrumb/index.js
```diff
@@ -2,15 +2,16 @@
...
 
-// Module dependencies
-import base from 'base';
+// Custom
+import 'base';
 
 // Module styles
 import './_breadcrumb.scss';
 
-export default {
```
jryanconklin: I'm curious what are the advantages by swapping approaches here?

source/design-system.js
```diff
@@ -1,17 +1,17 @@
...
+import * as breadcrumb from 'atoms/breadcrumb';
+import * as button from 'atoms/button';
+import * as image from 'atoms/image';
+import * as icon from 'atoms/icon';
+import * as listItem from 'atoms/list-item';
+import * as card from 'molecules/card';
+import * as jumbotron from 'molecules/jumbotron';
 
-// Plain ol' object in ES6 object shorthand syntax
-const designSystem = {
```
jryanconklin: If we're dropping use of `designSystem` here, will we able to update the `ds`' above back to `designSystem` without the console errors you were seeing?

source/design-system.js
```diff
@@ -1,17 +1,17 @@
...
 
-import breadcrumb from 'atoms/breadcrumb';
-import button from 'atoms/button';
-import image from 'atoms/image';
-import icon from 'atoms/icon';
-import listItem from 'atoms/list-item';
-import card from 'molecules/card';
-import jumbotron from 'molecules/jumbotron';
+import * as breadcrumb from 'atoms/breadcrumb';
+import * as button from 'atoms/button';
```
jryanconklin: Considering how simple our templates for `js` behavior in components is, does it make sense to do a wildcard import? Or are you trying to future-proof these?

webpack.shared.config.js
```diff
@@ -26,7 +26,8 @@ module.exports = {
         test: /\.(sass|scss)$/,
         include: [
           path.resolve(__dirname, 'source'),
-          path.resolve(__dirname, 'drupal'),
+          path.resolve(__dirname, 'app-drupal'),
+          path.resolve(__dirname, 'app-pl'),
```
jryanconklin: You forgot `app-grav`, but I'm sure Critter will forgive you. In all seriousness, I think this is great.

README.md
illepic: @jryanconklin, we're definitely going to find an intensely unique token name to make find/replace easier for folks. (thundersnow! JK. Kind of.)

source/_patterns/01-atoms/breadcrumb/index.js
illepic: All js files are "objects" from a module perspective. Our old way of doing this was essentially building an object in an "object", just to export. ES6 modules make this much simpler with the new structure.

source/design-system.js
tjheffner: this wildcard import ends up with the same result as the previous syntax. with how the js files are now broken up into discrete exports instead of one total object, we want to be sure we have the entire thing when we pull it into the full design system

tools/tasks/gulp-sass2json.js
```diff
@@ -1,105 +0,0 @@
-const path = require('path');
```
tjheffner: rip 🎉

source/design-system.js
illepic: This is how we get an entire component into a var, including all it's exported functions. So we can now throw around button and have a safe assumption it will always have `.enable()`, `.disable()` etc

source/design-system.js
illepic: CORRECT! I'll make this change.

webpack.shared.config.js
illepic: (@mikecrittenden). If we say Grav three times in a dark bathroom, staring into a mirror, he appears behind you with opinions about markdown-based, database-less CMS.

source/design-system.js
illepic: This is updated now at implementation in `app-drupal` and `app-pl`.

illepic:
@tjheffner, we're going to move `tempates/`, and the drupal yaml files into `app-drupal/` in a following PR. THAT is technically the theme and Drupal shouldn't care that the theme is just one folder deeper.

source/design-system.js
jryanconklin: Woot!

webpack.shared.config.js
mikecrittenden: IT'S THE DX OF A SSG WITH THE POWER OF A CMS

PR #167, opened 2017-11-10
Move drupal theme fully into app-drupal/

illepic:
The drupal theme now lives fully in `app-drupal/`. All paths and tooling for namespaces have been updated to the new path. Everything appears to work great.


app-drupal/index.js
```diff
@@ -19,9 +19,10 @@ _.forEach([1, 2, 3, 4], (num) => {
 });
 
 Drupal.behaviors.designSystem = {
-  attach(context, settings) {
-    designSystem.forEach((component) => {
-      component.enable(context, settings);
+  attach($context, settings) {
+    _.forEach(designSystem, (component) => {
+      console.log(component.name);
```
grayside: Is name enough context to make sense of the logging?

app-drupal/patternlab.info.yml
```diff
@@ -0,0 +1,42 @@
+name: Pattern Lab Starter
+type: theme
+description: A theme with Pattern Lab
```
grayside: The name and description seems out of date. I presume it's pending on the rename?

illepic: Definitely gets some love once we nail the name.

app-drupal/index.js
illepic: Yep, this is just not-production clarity for new folks coming in to see how this works.

illepic:
@grayside , not quite. No build steps, no symlinks.

Drupal has no problem finding a theme one folder deeper than `sites/all/themes/themename`. Our theme just now lives at `sites/all/themes/app-drupal`. This is the directory that will hold `.info.yml`, `.theme`, `.html.twig`, and `.libraries.yml` files.

app-drupal/templates/block--system-branding-block.html.twig
```diff
@@ -14,10 +14,19 @@
...
 
+
 {% block content %}
-  {% include "@molecules/branding.twig"
-    with {
-      "url": path('<front>')
-    }
-  %}
+
+  <h1>i r branding nerp</h1>
```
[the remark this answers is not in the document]
illepic: @jryanconklin is up in this file and it will be immediately revamped when his incoming PR hits!

app-drupal/templates/node--teaser.html.twig
```diff
@@ -63,13 +63,14 @@
...
-{% embed "@molecules/caption-block.twig"
-  with {
-    "title": label
-  }
-%}
-  {% block body %}
-    {{ content }}
-  {% endblock %}
-{% endembed %}
+{#{% embed "@molecules/caption-block.twig"#}
```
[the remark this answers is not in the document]
illepic: Roger that, @jryanconklin's work revamps all this.

grayside:
In Drupal 8, in a typical outrigger + GDT setup, source code lives in src/* and built/upstream code lives in build/html. The theme skirts this a bit, having it's own prepared & upstream code inside it's own directory under src. So What I'm trying to understand is which paths we want to take.

Regardless of those details, it sounds like you are encouraging this to be thought of as a nested theme, rather than as a design system embedded in another project that produces a theme. (The "Generated Theme" mode)

illepic:
@grayside, at this stage, having the very hard delineation between `thundersnow/source` and `thundersnow/app-drupal` is the biggest step toward eventually COMPLETELY separating them out. I think that's already a huge brain shift that I'll be on the hook for for years ;)

Stage 2 is breaking them out and having to make changes to overall project build processes.

PR #170, opened 2017-11-10
Move PL into the pl-app folder where it is more suitable

illepic:


illepic:
Updating shortly with the path changes.

.eslintignore
```diff
@@ -1,2 +1,7 @@
 # Ignore any path/file using .gitignore syntax, ie:
 # js/vendor
+
+dist/
+apps/pl/pattern-lab
+# Ignore yeoman new-component for now
+tools/new-component
```
tjheffner: new line

illepic: GAAARRRRRGGGG

config.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
+  PATH_PL: './apps/pl',
+  PATH_DRUPAL: './apps/drupal',
+  PATH_SOURCE: './source',
+};
```
tjheffner: So I know we moved configs to be explicit in the gulpfile, are you thinking some of those move back into here? thinking specifically of the namespacing & dev-server config. I am leaning no still, because I like them explicit right where they're used, so maybe.... just adding a comment about what makes this config.js special (i.e. because it's used in gulpfile & webpack.config, so it needs to be global)

illepic: Yep, I'll make a note that these are shared things between multiple files, not magic features config.

PR #175, opened 2017-11-13
Feature/v10/plp 30 drupal block

jryanconklin:
This PR adds:
* Branding molecule and drupal branding block
* Theme suggestions for Block Type and View Mode
* Drupal Card Block Example

On the card block example, I've left the variables open since these could be set via Drupal giving editors control.


apps/drupal/index.js
```diff
@@ -12,7 +12,7 @@ require('./scss/_drupal-styles.scss');
 console.log('drupal-theme ran');
 
 // Verifying that we have $
-$('h1').html('hello from jquery');
+// $('h1').html('hello from jquery');
```
illepic: Go ahead and nuke this + my silly foreach loop below.

apps/drupal/logo.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="150" height="200" viewBox="0 0 150 200"><path fill="none" stroke="#1A1A1A" stroke-width="10" stroke-miterlimit="10" d="M142.99 125.89c0 37.55-30.44 67.99-67.99 67.99S7.01 163.44 7.01 125.89 75 8.82 75 8.82s67.99 79.52 67.99 117.07z"/><path fill="#2AA8E0" d="M126.8 125.89c0 28.61-23.19 51.8-51.8 51.8-28.61 0-51.8-23.19-51.8-51.8C23.2 97.28 75 36.7 75 36.7s51.8 60.58 51.8 89.19z"/><ellipse fill="#FF7900" cx="33.1" cy="66.64" rx="23.08" ry="23.95"/></svg>
```
illepic: Drupal should be consuming the logo from `source/`.

source/_patterns/02-molecules/branding/_branding.twig
```diff
@@ -0,0 +1,35 @@
...
+#}
+
+{% set site_logo = {
+  svgpath: pl ? '@molecules/branding/logo.svg' : site_logo,
+} %}
+
+<aside class="branding">
+  {% if site_logo %}
+    <div class="branding__site-logo">
+      <a href="{{ (pl) ? '#' : url }}" title="{{ 'Home'|t }}" rel="home">
```
illepic: I think we can do this better.

source/_patterns/02-molecules/branding/_branding.twig
```diff
@@ -0,0 +1,35 @@
...
+ * url: String: the url the block directs to.
+ * site_logo: String: the logo provided by the app.
+ * site_name: String: the site name provided by the app.
+ * site_slogan: String: the site slogan provided by the app.
+ *
+ */
+#}
+
+{% set site_logo = {
+  svgpath: pl ? '@molecules/branding/logo.svg' : site_logo,
```
illepic: Let's revisit this.

source/_patterns/02-molecules/branding/demo/brandings.yaml
```diff
@@ -0,0 +1,3 @@
+url: 'http://0.0.0.0:8080/pl/'
+site_name: 'Site Name'
+site_slogan: 'A Clever Slogan'
```
illepic: Install an editorconfig plugin to your IDE to automatically use our project config. Will auto-add line endings.

source/_patterns/02-molecules/branding/index.js
```diff
@@ -0,0 +1,22 @@
+/**
+ * Branding
+ */
+
+// Module dependencies
+import 'base';
+import 'atoms/image';
+import 'atoms/button';
```
illepic: Does `branding` need button? I don't see a button atom used within it.

source/_patterns/02-molecules/branding/index.js
```diff
@@ -0,0 +1,22 @@
...
+
+// Pull in system-wide images
+import './logo.svg';
```
illepic: Why duplicate the logo within image?

source/_patterns/02-molecules/branding/logo.svg
```diff
@@ -0,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" width="150" height="200" viewBox="0 0 1 1">
```
illepic: See above comment about duplicate logo.

source/_patterns/02-molecules/branding/index.js
jryanconklin: It doesn't as-is, no. Had left in there as a part of the branding test you'd recommended originally, e.g., add a button to see if it works. I'll remove.

source/_patterns/02-molecules/branding/_branding.twig
anniegreens: Let's use the method I posted in the flow a week or so ago. It basically sets up all variables for demo purposes in PL if the variable array is not already being passed (such as from a source that includes/embeds/extends the pattern like a page that wants to pass its own data and not Faker/Lorem Ipsum).

```
{% if pl and not brandingBlock %}
  {% set brandingBlock = {
    variable1: value1,
    variable2: value2,
  } %}
{% endif %}
```

We then also have data encapsulation, by namespacing our variables. This avoids headaches later when passing variables to included patterns at a higher level. You'd then use these as:

`{{ brandingBlock.variable1 }}` etc.

source/_patterns/02-molecules/branding/_branding.twig
anniegreens: See above suggestion.

source/_patterns/02-molecules/branding/logo.svg
jryanconklin: Oh! It wasn't really a duplicate logo. They were separate logos to solve using a graybox for PL vs. the "real" one for Drupal. Probably should simply set the site logo and let it be determined by the file?

source/_patterns/02-molecules/branding/_branding.twig
jryanconklin: Awesome, on it!

jryanconklin:
@illepic @anniegreens I'm still working on this but wanted some feedback. I think this works for the PL demo, we're able to display both versions in the brandings demo. It's a little awkward bringing it over to Drupal, only the SVG inline version works currently.

It's more complex but it manages both having SVGs and ImgSrc taking the approach we took on whether the container exists. I had to stuff the inline value inside the brandingBlock in order to run the if statement.

apps/drupal/patternlab.theme
```diff
@@ -4,3 +4,45 @@
...
  */
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function patternlab_suggestions_block_alter(array &$suggestions, array $variables) {
+  // Provide suggestion for block templates by custom block type.
+  if (!empty($variables['elements']['content']['#block_content'])) {
+    $block = $variables['elements']['content']['#block_content'];
+    // Add `block--BLOCK-TYPE.html.twig`
```
illepic: We're super sure that Drupal 8 doesn't do this already?

apps/drupal/patternlab.theme
```diff
@@ -4,3 +4,45 @@
...
+function patternlab_preprocess_block(&$variables) {
+  $id = $variables['plugin_id'];
+
+  if ($id === 'system_branding_block') {
+    $logo_default = theme_get_setting('logo.use_default');
+    $logo_uri = theme_get_setting('logo.url');
+    $file_type = substr($logo_uri, -3);
+
+    // Set the logo to PL if default.
+    if ($logo_default === 1) {
```
illepic: Let's just `if ($logo_default)`

jryanconklin:
@illepic Updated! I made the changes we discussed earlier today. Spent a bit of time on the preprocess to handle:
* If it's our default logo (use PL)
* If the user supplies a custom logo that's an SVG
* If the user supplies a custom logo that's not a regular image (or default lightning image)

The only thing this doesn't handle is if the logo is set within `src/themes/pattern-lab-starter/apps/drupal/templates/block/block--system-branding-block.html.twig` where a condition for the other is set in the preprocess. This causes a duplicate logo.

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -8,25 +8,17 @@
...
-
 {% block content %}
 
-  <h1>i r branding nerp</h1>
-
-  {% include '@atoms/button/_button.twig' with {
-    button_size: 'lg',
-    button_text: 'hello',
-  } %}
+  {% include "@molecules/_branding.twig" %}
```
illepic: No `with{}` here?

source/_patterns/02-molecules/branding/_branding.twig
```diff
@@ -0,0 +1,44 @@
...
+ * site_slogan: String: the site slogan provided by the app.
+ *
+ */
+#}
+
+<aside class="branding">
+  <div class="branding__site-logo">
+    <a href="{{ url }}" title="{{ site_name }}" rel="home">
+
+      {% if logo_inline_svg %}
```
illepic: The code is out of alignment here and below with the other if block.

apps/drupal/patternlab.theme
jryanconklin: Triple double sure!

apps/drupal/templates/block/block--system-branding-block.html.twig
jryanconklin: I can add it for demo purposes, but the logo atom itself is set via the preprocess so it's not really necessary. Plus taking into account the possibility of a duplicate logo I left it out.

jryanconklin:
@illepic Updated to handle with in Twig. Also added unset to the preprocess to be super double specific.

jryanconklin:
@illepic Updated! Moved Branding from molecule => atom.

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -8,25 +8,19 @@
...
-
 {% block content %}
 
-  <h1>i r branding nerp</h1>
-
-  {% include '@atoms/button/_button.twig' with {
-    button_size: 'lg',
-    button_text: 'hello',
+  {% include "@atoms/_branding.twig" with {
+    logo_inline_svg: '@atoms/image/logo.svg',
```
illepic: Hey, since this is the Drupal block, bring over the if uploaded checks just like in the demo of this in PL.

apps/drupal/templates/block/block--basic--card.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ * @see @molecules/_card.twig for full list of variables.
+ *
+ * text_card for example purposes.
+ * Set variables directly in template or as content editor created fields.
+ *
+ */
+#}
+
+{% set block_card = {
+    card_width: card_width ? card_width : 100,
```
illepic: Update all these `thing ? thing : defaultValue` to `thing ?: defaultValue`

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -8,25 +8,25 @@
...
 {% block content %}
 
-  <h1>i r branding nerp</h1>
-
-  {% include '@atoms/button/_button.twig' with {
-    button_size: 'lg',
-    button_text: 'hello',
+  {# Use logo to explicitly override values from preprocess. #}
+  {% set logo = {
+    logo_svg_inline: logo_inline_svg ? logo_inline_svg : '',
```
illepic: See above about `?:`.

jryanconklin:
@illepic Updated!

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -15,16 +15,15 @@
...
  */
 #}
 
 {% block content %}
 
   {# Use logo to explicitly override values from preprocess. #}
   {% set logo = {
-    logo_svg_inline: logo_inline_svg ? logo_inline_svg : '',
-    logo_img_src: logo_img_src ? logo_img_src : ''
+    logo_svg_inline: logo_inline_svg ?: logo_inline_svg,
```
illepic: This doesn't make sense.

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -15,16 +15,15 @@
...
+    logo_img_src: logo_img_src ?: logo_img_src
```
illepic: Neither does this.

jryanconklin: Ah! Misread your last comment. Updating.

jryanconklin:
@illepic Updated!

apps/drupal/templates/block/block--system-branding-block.html.twig
```diff
@@ -22,8 +22,8 @@
 
   {# Use logo to explicitly override values from preprocess. #}
   {% set logo = {
-    logo_svg_inline: logo_inline_svg ?: logo_inline_svg,
-    logo_img_src: logo_img_src ?: logo_img_src
+    logo_svg_inline: logo_inline_svg,
```
illepic: Trying to verify the `logo_svg_inline` vs `logo_inline_svg` naming weirdness here, while lining it up with the var name mentioned in the docblock at top of file.

jryanconklin:
@illepic Updated!

PR #176, opened 2017-11-13
[HOLD] epic/v10 into master

illepic:
Just putting this up early so we have ongoing stats as to the enormity of `epic/v10`.


illepic:
We're doing it. BOMBS AWAY.

PR #182, opened 2017-11-28
Feature/v10/plp 29 drupal page

jryanconklin:
This PR adds:
* Drupal Configuration: Block and Regions by Default in Theme
* Breadcrumb Template / Implementation
* Header / Footer Template / Implementation

**Important Note**: If you're working with this in Drupal, you'll need to uninstall the theme, clear cache and reinstall.


apps/drupal/patternlab.theme
```diff
@@ -39,6 +39,29 @@ function patternlab_preprocess_block(&$variables) {
     } else {
       $variables['logo_img_src'] = $logo_url;
     }
+  }
+
+  if ($id === 'system_powered_by_block') {
```
grayside: Immediate thought here: should we explicitly prefer block-specific preprocess where they don't have common needs? That would help promote discoverability I think. E.g., `template_preprocess_block_system_branding_block()`.

apps/drupal/patternlab.theme
```diff
@@ -39,6 +39,29 @@ function patternlab_preprocess_block(&$variables) {
...
 
+/**
+ * Implements hook_preprocess_region().
+ */
+function patternlab_preprocess_region__header(&$variables) {
+  $blocks = [
+    'branding' => 'patternlab_sitebranding',
+  ];
+  foreach ($blocks as $key => $value) {
+    // Begins with value.
```
grayside: I understand it begins with value, but I'm really not clear what the goal is of this code snippet. Stronger what & why comment?

apps/drupal/templates/block/block--system-menu-block.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ * - title_suffix: Additional output populated by modules, intended to be
+ *   displayed after the main title tag that appears in the template.
+ *
+ * Headings should be used on navigation menus that consistently appear on
+ * multiple pages. When this menu block's label is configured to not be
+ * displayed, it is automatically made invisible using the 'visually-hidden' CSS
+ * class, which still keeps it visible for screen-readers and assistive
+ * technology. Headings allow screen-reader and keyboard only users to navigate
+ * to or skip the links.
+ * See http://juicystudio.com/article/screen-readers-display-none.php and
```
grayside: Some tools will extract `@see` tag. Insert a newline above, use one @see per url, remove other words to use that structure.

apps/drupal/templates/layout/page.html.twig
```diff
@@ -0,0 +1,97 @@
...
+          {{ page.sidebar_second }}
+        </aside>
+      {% endif %}
+
+    </section>
+  </main>
+
+</div>{# /.layout-container #}
+
+  {{ page.footer }}
```
grayside: Extra leading space?

source/design-system.js
```diff
@@ -6,20 +6,23 @@
 import * as branding from 'atoms/branding';
 import * as breadcrumb from 'atoms/breadcrumb';
 import * as button from 'atoms/button';
+import * as grid from 'atoms/grid';
```
grayside: To these have a particular order?

jryanconklin: Yeah, thought just putting them in alpha order by type, e.g., atoms, molecules, etc.

apps/drupal/templates/block/block--system-menu-block.html.twig
jryanconklin: I'll probably pull this, I'd been working on it based on Anne's original work but the initial layout is going a different direction. But good to know, thank you!

apps/drupal/patternlab.theme
jryanconklin: Fair, I'll swap this for these two blocks!

apps/drupal/patternlab.theme
jryanconklin: I'll expand this out.

source/_patterns/01-atoms/breadcrumb/_breadcrumb.twig
```diff
@@ -14,12 +14,12 @@
 <nav aria-label="breadcrumb" role="navigation">
   <ol class="breadcrumb">
     {% set break = FALSE %}
-    {% for item in breadcrumb_items if not break %}
+    {% for item in breadcrumb if not break %}
```
tjheffner: we can assume that the active breadcrumb will be the last breadcrumb in the list so this & the .yml can be cleaned up a bit with [twigs loop helpers](https://twig.symfony.com/doc/2.x/tags/for.html#the-loop-variable). both break and breadcrumb_active can be nixed.

`{% if loop.last %} active breadcrumb {% else %} normal breadcrumb {% endif %}`

will also need to add a small preprocess inside of patternlab.theme to fully populate the breadcrumb array (without it, I'm only seeing "Home" render as the active link, regardless of page)

```
/**
 * Implements hook_preprocess_breadcrumb().
 */
function patternlab_preprocess_breadcrumb(&$variables){
  if(($node = \Drupal::routeMatch()->getParameter('node')) && $variables['breadcrumb']){
    $variables['breadcrumb'][] = array(
      'text' => $node->getTitle()
    );
  }
}
```

source/_patterns/03-organisms/header/_header.scss
```diff
@@ -0,0 +1,12 @@
+// All component Sass needs non-printing base config
+@import '../../00-base/config';
+
+// To change default bootstrap jumbotron variables, SEE ~bootstrap/scss/variables
+// Add bootstrap variables here
+
+//$jumbotron-padding: some_override_value;
+//$jumbotron-bg: some_override_value;
+
+@import '~bootstrap/scss/jumbotron';
```
illepic: If `jumbotron` already exists as a pattern, do **not** included it here. Just having the dependency in your `header/index.js` pulls over jumbotron styles. So, remove this line and just have your custom header sass if needed.

source/_patterns/03-organisms/header/demo/headers.twig
```diff
@@ -0,0 +1,6 @@
+<header role="banner">
+  {% embed '@organisms/header/_header.twig' %}
+    {% block branding %}
```
illepic: Doing this to wipe out the branding for the header demo?

source/_patterns/03-organisms/header/index.js
```diff
@@ -0,0 +1,17 @@
+/**
+ * Header
+ */
+
+// Module dependencies
+import 'base';
+
```
illepic: Needs `jumbotron` right?

source/_patterns/03-organisms/header/demo/headers.twig
jryanconklin: Thought here was to leave it following the pattern we use elsewhere to suggest it can be overridden. That said, I'm still not totally understanding the little differences between include, extends and embed (other than embed is combining include + extends?). Plan on putting some time into reading about it today since a lot of this branch could have been easier with that knowledge.

source/_patterns/03-organisms/header/_header.scss
jryanconklin: Ah! My bad, we had a bit chat about this early on and time's made that a bit muddy. Removing ASAP.

jryanconklin:
@illepic Updated! See additional comment on `source/_patterns/03-organisms/header/demo/headers.twig`.

source/_patterns/03-organisms/header/demo/headers.twig
illepic: Having the empty block here in this implementation blows away whatever lives by default in this twig blog within the `_header.twig` file (which is the branding atom).

jryanconklin:
@illepic Updated!

apps/drupal/particle.theme
```diff
@@ -0,0 +1,77 @@
...
+function particle_preprocess_region__header(&$variables) {
+  // Blocks contains list of blocks we want to call individually in Twig.
+  // Example, {{ content }} => {{ branding }}
+  $blocks = [
+    'branding' => 'particle_sitebranding',
+  ];
+  foreach ($blocks as $key => $value) {
+    $pattern = '/^' . $value . '/';
+    foreach ($variables['elements'] as $el_key => $el_value) {
+      if(preg_match($pattern, $el_key)) {
```
illepic: Let's talk about what's happening here. Regex at this stage is a smell that makes me uncomfortable.

source/_patterns/03-organisms/header/_header.scss
```diff
@@ -0,0 +1,10 @@
+// All component Sass needs non-printing base config
+@import '../../00-base/config';
+
+// To change default bootstrap jumbotron variables, SEE ~bootstrap/scss/variables
+// Add bootstrap variables here
+
+//$jumbotron-padding: some_override_value;
```
illepic: Delete mentions of jumbotron

jryanconklin:
@illepic Updated!

apps/drupal/templates/layout/page.html.twig
illepic: It's a child of `<footer>`.

apps/drupal/templates/layout/region--footer.html.twig
```diff
@@ -13,17 +13,13 @@
  */
```
illepic: @jryanconklin: we don't need a footer region template. If there are wrapper classes, place them up in page.

apps/drupal/templates/layout/region--header-branding.html.twig
```diff
@@ -13,18 +13,13 @@
  */
```
illepic: We don't need a template for this region.

source/_patterns/02-molecules/jumbotron/_jumbotron.twig
```diff
@@ -26,17 +26,17 @@
     <div class="container">
   {% endif %}
 
-  {% block branding %}
+  {% block header_branding %}
```
illepic: I'd very much like to name the jumbotron regions something quite different from the drupal regions. Call these `jumbotron_branding` and `jumbotron_cta`. This will make a clearer distinction between what is the design system and what is drupal stuff.

jryanconklin:
@illepic Updated!

Issue #169, opened 2017-11-10
Update twig demos that reference baseScssVars


illepic:
We removed a large baseScssVars var, a few twig demos still reference it. They need to be updated to our new sass2json method.

Issue #157, opened 2017-11-07
Consolidate, cleanup, throwaway files in _data/


illepic:
_data folder needs some serious attention. All unused info should be tossed.

tjheffner:
cleanup started here, still needs work/evaluation of what data is actually being used. https://github.com/phase2/particle/pull/257

both .json files stay [auto-generated]
device_widths.yml stays
imageStyles.yml stays [we should be consistent on casing, so deviceWidths]

tables/texts/times are all nice helpers but not sure if anything actually uses them

data.yml needs cleaned up [should keep global pl nav/footer, some other things, but can be reduced heavily]

listItems needs simplified/cleaned/nuked if nothing uses it

Issue #155, opened 2017-11-07
Provide web-based iconfont example


illepic:
Let's pull in FontAwesome and demo a few fontawesome icons. Handle via cdnjs:

https://cdnjs.com/libraries/font-awesome

Issue #125, opened 2017-10-16
All patterns should follow the new "demo" structure


illepic:
Nearly every pattern should follow this folder structure. 

```
02-molecules/
  |-- card
    _card.twig
    |-- demo
      card-large.twig
      card-large.yml
      card-dark.twig
      card-dark.yml
      card-no-image.twig
      card-no-image.yml
```

This allows purely demonstration patterns to live in a folder than can be optionally removed from deployment to production. The idea being that anything in `demo` will **not** be included in any other pattern.

What if you **want** to use a pre-loaded pattern usable elsewhere? Simple, just include the re-usable patterns at the top pattern level with some side-car data, ie:

```
02-molecules/
  |--card
    _card.twig
    card-large.twig # This embeds _card.twig with some hardcoded variant values
    card-large.yml
    |--demo
      card-dark.twig
      card-dark.yml
      card-no-image.twig
      card-no-image.yml
```

tjheffner:
https://github.com/phase2/particle/pull/257

PR #184, opened 2017-12-04
general cleanup

tjheffner:
what i could pull back from msk:

- cleaned up 00-base a little bit, 
- moves breakpoints & spacing into the scss2json pipeline, 
- added target="_blank" in a few spots, 
- added fontawesome icons
- started to clean up the templates & homepage pages, still needs some work to fit the out-of-box bootstrap paradigm

should hit some parts/all of these issues:
https://github.com/phase2/particle/issues/177
https://github.com/phase2/particle/issues/169
https://github.com/phase2/particle/issues/157
https://github.com/phase2/particle/issues/125 [this issue desc. should get copied into docs]
https://github.com/phase2/particle/issues/155


source/_data/scssVariables.json
```diff
@@ -10,6 +10,18 @@
     "dark": "rgb(52, 58, 64)",
     "test": "rgb(255, 192, 203)"
   },
+  "scssBreakpoints": {
```
jryanconklin: Nice. If we update this in the grid component, this will update correct?

source/_patterns/00-base/_base.scss
```diff
@@ -20,8 +20,13 @@
...
+// Font Awesome icons
+$fa-font-path: '../../fonts';
+@import '~font-awesome/scss/font-awesome';
+
 // Custom CSS-generating output
-@import 'type';
-@import 'forms';
-@import 'tables';
-@import 'transitions';
+@import 'printing/type';
```
jryanconklin: This is much clearer. Thank you!

source/_patterns/00-base/_variables.scss
```diff
@@ -0,0 +1,37 @@
+///
```
jryanconklin: Are we going to aim to have this more fully documented on the front end docs project?

source/_data/scssVariables.json
tjheffner: these values are coming from bootstrap's $grid-breakpoints(). if needed, they would get overridden probably inside of _variables.scss, similar to how we are able to override $theme-colors()

source/_patterns/00-base/_variables.scss
tjheffner: yes. this is mostly a companion to the large comment in _base.scss, the other "main" sass file, to hopefully really underscore the printing/nonprinting difference. that distinction should/will be covered thoroughly in the FE docs

source/_patterns/00-base/_variables.scss
```diff
@@ -0,0 +1,37 @@
...
+@import 'non-printing/colors';
+@import 'non-printing/mixins';
+
+// (Other non-printing .scss), ie font families, spacing, etc
+$spacing: (
+  s: 5px,
+  m: 25px,
+  l: 50px,
+);
+
```
illepic: Comment right here that this is where we can override bootstrap variables (that aren't just overriden in a component later on).

PR #185, opened 2017-12-05
Feature/v10/plp 28 drupal article

jryanconklin:
This PR aims to wrap up the Drupal side of Article as well as several additional updates. It includes importing Alert and adding a template to show off a demo of frontpage.


apps/drupal/templates/field/image.html.twig
```diff
@@ -0,0 +1,25 @@
...
+ *   We're using as a class below. Set you style in Drupal to a
+ *   Bootstrap style for automatic styling.
+ *   See https://getbootstrap.com/docs/4.0/content/images/
+ *
+ * @see template_preprocess_image()
+ */
+#}
+
+{% set image_classes = [
+  'm-3',
```
illepic: eeeeeh, I don't think this should go here.

apps/drupal/templates/field/image.html.twig
```diff
@@ -0,0 +1,25 @@
...
+  'img',
+  'img-responsive',
+  style_name|clean_class,
+  'float-right'
```
illepic: Same with this. This will affect ALL image fields, not just on this view mode/template. Need to move these positional things to a different hook.

apps/drupal/templates/misc/status-messages.html.twig
```diff
@@ -0,0 +1,56 @@
...
+
+{% set status_heading = {
+  'status': 'Status message'|t,
+  'error': 'Error message'|t,
+  'warning': 'Warning message'|t,
+  'info': 'Informative message'|t,
+  }
+%}
+
+{% for type, messages in message_list %}
```
illepic: Are we printing all out on the page as part of the demo in Drupal?

apps/drupal/templates/navigation/menu--account.html.twig
```diff
@@ -0,0 +1,48 @@
...
+ *   - in_active_trail: TRUE if the link is in the active trail.
+ */
+#}
+{% import _self as menus %}
+
+{#
+  We call a macro which calls itself to render the full tree.
+  @see http://twig.sensiolabs.org/doc/tags/macro.html
+#}
+{{ menus.menu_links(items, attributes, 0) }}
```
illepic: :HEADEXPLODE:

source/_patterns/01-atoms/alert/_alert.scss
```diff
@@ -0,0 +1,9 @@
+// All component Sass needs non-printing base config
+@import '../../00-base/config';
```
illepic: Update this with @tjheffner 's new structure.

source/_patterns/01-atoms/alert/_alert.twig
```diff
@@ -0,0 +1,52 @@
...
+  'mb-0',
+  'alert',
+  'alert-' ~ status_classes[type],
+  'alert-dismissible',
+  'fade',
+  'show',
+] | sort | join(' ') | trim %}
+
+{% if type == 'error' %}
+  {% set alert_degree = 'role=alert' %}
```
illepic: {% set alert_degree = ( type == 'error' ? 'role=alert' : 'aria-live=polite) %}` or whatever the cleanest shorthand is in PHP/Twig.

source/_patterns/03-organisms/card-grid/_card-grid.twig
```diff
@@ -0,0 +1,26 @@
...
+  This is provided to show you an example of our Cards in grid in Pattern Lab.
+  Additional implementations are app-specific and should be handled on a case-by-case basis.
+
+  see https://getbootstrap.com/docs/4.0/components/card/ for more details
+#}
+
+<div class="row">
+  {% for item_card in card_grid_rows %}
+    {% set item_card = {
+      card_cols: 'col-sm-12 col-md-6 col-lg-4',
```
illepic: So, I know that cards have this concept of being their own columns in bootstrap 4, but why aren't we doing this with wrapper items instead? Are we gaining something by junking up the card with these extra classes?

apps/drupal/templates/field/image.html.twig
madhaze: Some of these look like very specific classes add to all images (m-3, float-right).  This file name should probably be updated to the field you are trying to affect here.

apps/drupal/templates/misc/status-messages.html.twig
jryanconklin: Yeah, but none really exist. I'd been adding them via preprocess to test to make sure it could handle them individually or in bulk. But really, this should handle any error we give it in any order.

source/_patterns/03-organisms/card-grid/_card-grid.twig
jryanconklin: I think I originally had a wrapper class on the `views-view-unformatted--frontpage.html.twig` template that applied these styles, but we'd chatted about having that template be a bit simpler. FWIW I think this is simpler / easier to read, but I certainly cede to your experience.

apps/drupal/templates/navigation/menu--account.html.twig
jryanconklin: This is direct from stable. I think these can be more extensively refactored once @rloos289 nav work is brought in.

jryanconklin:
@illepic Updated!

apps/drupal/templates/misc/status-messages.html.twig
illepic: @jryanconklin : let's do this: in the .theme file, fire off a warning, error, and info message. Maybe have it fire on every single page load just to show it off, informing users to comment remove it.

apps/drupal/templates/navigation/menu--account.html.twig
illepic: Since this is straight from Stable, let's get some comments about the Particle-specific additions to the code, and what those additions do.

illepic: Noting here that this is pretty low touch, dont' need a lot of detail

jryanconklin:
@illepic Updated!

source/_patterns/00-base/non-printing/_colors.scss
```diff
@@ -2,6 +2,8 @@
 
 // Bootstrap overrides
 // Add colors to $theme-colors var here. Find other vars to override at ~bootstrap/scss/_variables.scss
+// Be sure to add new colors in quotes.
 $theme-colors: (
-  test: pink,
+  "test": pink,
```
illepic: Single tick instead of double.

source/_patterns/04-templates/homepage/homepage.twig
```diff
@@ -0,0 +1,43 @@
+{#
+  Homepage Template
+
+  This is the template to display our homepage!
+
+#}
+
+{% embed '@atoms/alert/_alert.twig' with alert_info %}
```
illepic: If not overriding twig blocks, use `include`.

webpack.grav.config.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Grav-specific webpack config
+ * This is merged over top of webpack.shared.config.js
+ */
+const path = require('path');
+const merge = require('webpack-merge');
+const shared = require('./webpack.shared.config');
```
illepic: I'd like to refactor these webpack configs to move the custom imports (`shared`) to the bottom with a space between custom imports and library imports.

jryanconklin:
@illepic Updated!

PR #187, opened 2017-12-11
Feature/v10/plp 13 bootstrap carousel

jryanconklin:
This PR adds Bootstrap Carousel with:
* Slide Only
* Controls
* Indicators
* Image Captions

### DIFF
--- a/source/_data/imageStyles.yaml
+++ b/source/_data/imageStyles.yaml
@@ -12,3 +12,6 @@ imageStyles:
     width: 800
   large:
     width: 1600
+  carousel:
+    width: 1110
+    height: 400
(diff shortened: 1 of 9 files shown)


source/_patterns/02-molecules/carousel/index.js
```diff
@@ -0,0 +1,26 @@
...
+import 'atoms/image';
+
+// Module styles
+import './_carousel.scss';
+
+export const name = 'carousel';
+
+export function disable() {}
+
+export function enable() {
```
illepic: Need `context` here.

jryanconklin:
@illepic Updated with context!

source/_data/imageStyles.yaml
```diff
@@ -12,3 +12,6 @@ imageStyles:
     width: 800
   large:
     width: 1600
+  carousel:
```
anniegreens: This seems like a strange image size/ratio. Can we better rely on a more standard width and then grab a corresponding height based on a 3:2 ratio or other standard?

PR #188, opened 2017-12-11
add font awesome

tjheffner:
this PR:

- adds font-awesome as an npm dependency
- includes local font files in source/
- demonstrates how to import an npm module's sass and overwrite variables for that sass locally
- updates icons demo to show font-awesome implementation also


source/_patterns/01-atoms/icon/scss/_icon.scss
```diff
@@ -3,6 +3,9 @@
 ///
 
 @import 'icon-mixins';
+@import '~font-awesome/scss/mixins';
+@import '~font-awesome/scss/variables';
```
tjheffner: just noting these files don't have printing output, totally fine to include here.

PR #191, opened 2017-12-12
PLP-26: Update Grid for Container Fluid

jryanconklin:
This PR updates Grid to allow for `container-fluid` or empty values (no grid) without changing how grid / layouts themselves work.

### DIFF
--- a/source/_patterns/01-atoms/grid/_grid-1-up.twig
+++ b/source/_patterns/01-atoms/grid/_grid-1-up.twig
@@ -3,10 +3,12 @@
   in a .container div. We leave this up to the implementor manually for flexibility.
 #}
 
-{% if container %}<div class="container">{% endif %}
-<div class="row">
-  <div class="col">
-    {% block column_1 %}{% endblock column_1 %}
+{% set container_class = container ? 'container' : (container_fluid ? 'container-fluid') %}
+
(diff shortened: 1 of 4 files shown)


source/_patterns/01-atoms/grid/_grid-1-up.twig
```diff
@@ -3,10 +3,12 @@
   in a .container div. We leave this up to the implementor manually for flexibility.
 #}
 
-{% if container %}<div class="container">{% endif %}
-<div class="row">
-  <div class="col">
-    {% block column_1 %}{% endblock column_1 %}
+{% set container_class = container ? 'container' : (container_fluid ? 'container-fluid') %}
```
illepic: How can we do this with twig filters? And if not twig filters then more use of `?:`.

jryanconklin: I wanted this to be explicit to avoid refactoring usages of our grid embeds elsewhere, since we have a ton of `with { container: true }` throughout. This way it works as:
* No value
* `with { container: true }`
* `with { container-fluid: true }`

I think it's ugly in this template for the payoff of it being easier to use. FWIW.

jryanconklin: @illepic ^^

PR #201, opened 2017-12-22
Feature/v10/plp 35 drupal navbar

jryanconklin:
This PR:
* Adds Search form / config for Views and Search module
* Themes Primary menu region for two blocks: main menu and particle-search(s)

This works in Drupal but I don't think it's completely ready for primetime. But I'm looking to get the conversation started. It's not really extending (yet) from Pattern Lab. Trying to think through how to break apart these elements to extend from PL rather than hardcoding the individual pieces.

Another thing to consider is I'm moving the `<nav>` wrapper element from the `menu--main.html.twig` file to the region file at `region--primary-menu.html.twig`. Doing this in order to make sure the Search form is inside `nav`.

### DIFF
--- a/apps/drupal/particle.theme
+++ b/apps/drupal/particle.theme
@@ -5,6 +5,8 @@
  * Functions to support theming in the Particle theme.
  */
 
+use Drupal\Core\Form\FormStateInterface;
+
 /**
  * Implements hook_theme_suggestions_HOOK_alter().
  */
--- a/apps/drupal/particle.theme
+++ b/apps/drupal/particle.theme
@@ -87,3 +87,39 @@ function particle_preprocess_page(&$variables) {
   }
 
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function particle_theme_suggestions_form_alter(array &$suggestions, array $variables) {
+  $id = $variables['element']['#id'];
+  if ($id == 'views-exposed-form-search-page' || $id == 'search-block-form') {
+    $suggestions[] = 'form__particle_search';
+  }
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function particle_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Switch method to allow additional cases for these forms.
+  // Let's be explicit because it's easier for newcomers.
+  $id = $form['#id'];
+  switch($id) {
+    case 'views-exposed-form-search-page':
+      // Clean the unwanted label.
+      unset($form['#info']['filter-search_api_fulltext']['label']);
+      // Update type and add Bootstrap classes.
+      $form['keywords']['#type'] = 'search';
+      $form['keywords']['#size'] = 15;
+      $form['keywords']['#title'] = t('Enter the terms you wish to search for.');
+      $form['keywords']['#title_display'] = 'invisible';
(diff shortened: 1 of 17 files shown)


apps/drupal/particle.theme
```diff
@@ -87,3 +89,39 @@ function particle_preprocess_page(&$variables) {
   }
 
 }
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function particle_theme_suggestions_form_alter(array &$suggestions, array $variables) {
+  $id = $variables['element']['#id'];
+  if ($id == 'views-exposed-form-search-page' || $id == 'search-block-form') {
```
illepic: Comment above this line for WHY and WHAT.

apps/drupal/particle.theme
```diff
@@ -87,3 +89,39 @@ function particle_preprocess_page(&$variables) {
...
+    $suggestions[] = 'form__particle_search';
```
illepic: Comment here that the file is drupal-only, located at `yaddayadda/themes/forms/form--particle-search.twig`

apps/drupal/particle.theme
```diff
@@ -87,3 +89,39 @@ function particle_preprocess_page(&$variables) {
...
+  }
+}
+
+/**
+ * Implements hook_form_alter().
+ */
+function particle_form_alter(&$form, FormStateInterface $form_state, $form_id) {
+  // Switch method to allow additional cases for these forms.
+  // Let's be explicit because it's easier for newcomers.
```
illepic: Move comments to above switch.

illepic: Also make explicit note at the top here why we have to two cases when both share the same twig file. (built in drupal search form has radically different render array structure than view form render array)

apps/drupal/particle.theme
```diff
@@ -87,3 +89,39 @@ function particle_preprocess_page(&$variables) {
...
+  $id = $form['#id'];
+  switch($id) {
+    case 'views-exposed-form-search-page':
+      // Clean the unwanted label.
+      unset($form['#info']['filter-search_api_fulltext']['label']);
+      // Update type and add Bootstrap classes.
+      $form['keywords']['#type'] = 'search';
```
illepic: Make comment that the form field name is `keywords` (this is very confusing before figuring this out)

apps/drupal/templates/form/container.html.twig
```diff
@@ -0,0 +1,28 @@
...
+ * - children: The rendered child elements of the container.
+ * - has_parent: A flag to indicate that the container has one or more parent
+     containers.
+ *
+ * @see template_preprocess_container()
+ */
+#}
+{%
+  set classes = [
+    'form-group',
```
illepic: Comments per class here.

apps/drupal/templates/form/form-element.html.twig
```diff
@@ -0,0 +1,95 @@
+{#
```
illepic: Comments on what was added here beyond default.

apps/drupal/templates/navigation/menu--main.html.twig
```diff
@@ -0,0 +1,65 @@
+{#
```
illepic: A few helper comments.

jryanconklin:
@illepic: Updated!

PR #209, opened 2017-12-28
Pa11y Updates to pa11y 5 JavaScript Interface

jryanconklin:
This PR:
* Removes pa11y from `package.json`
* Removes Gulp pa11y tasks
* Updates README with download instructions
* Includes `pa11y.js` and `npm run pa11y` to scripts

This is the first half of the work here. We still need to resolve:
* Overlong pa11y header in console
* Whitespace from pa11y reporter
* Better handling of URLs array


tools/pa11y.js
```diff
@@ -0,0 +1,36 @@
+const pa11y = require('pa11y');
+const reporter = require('pa11y-reporter-cli'); // As pa11y 5 stabilizes, we can pull this off pa11y
+
+// @TODO dynamically retrieve pl data from config.js.
+// An Array of URLs for pa11y to test.
+const urls = [
```
illepic: For shits and giggles, let's:

```js
const url = require('url');

const plPath= 'http://0.0.0.0:8080/pl';
const testPaths= [
  '/patterns/00-protons-demo-borders/00-protons-demo-borders.html',
  // ...
];
// ...
const pallyPaths = testPaths.map(path => pa11y(url.resolve(plPath, path), options));
// ...
```
This brings over a couple of best practices around DRY, using proper url concatenation using built-in node functions

jryanconklin:
@illepic, okay, this is updated but `pa11y.js` isn't quite working as of yet. There's a reference error:
```
Evaluation failed: ReferenceError: _runPa11y is not defined
    at options (<anonymous>:3:3)
```
I think it's at line 37, because it's showing that `reporter.results(result)` is an unresolved function. Console logging a few versions of `pa11yPaths` it does look like we're sending it what Promise.all needs, e.g., it's running pa11y on the correct url paths which leaves `options` as the problem.

illepic:
@jryanconklin , blerp, I was wrong about what [url.resolve()](https://nodejs.org/api/url.html#url_url_resolve_from_to) actually does. It's not a "smart concatentator" like `path.resolve()`. You'll notice it's actually removing the `/pl/` part of the URL. 

Let's just use simple ES6 template string literals to get the job done here.

jryanconklin:
@illepic Updated!

tools/pa11y.js
```diff
@@ -26,12 +26,12 @@ const options = {
...
 
 /**
- * Returns an array of URLs with pa11y options.
+ * Returns an array of URLs with pa11y options. Join url via template literals.
  * @param {Array} testPaths - the paths we want to test.
  * @param {Array} options - the pa11y options we evaluate against.
  * @return {Array} pa11yPaths - a pa11y array to operate on.
  */
-const pa11yPaths = testPaths.map(path => pa11y(url.resolve(plPath, path), options));
+const pa11yPaths = testPaths.map(path => pa11y(`${plPath}${path}`, options));
```
illepic: Let's change the `path` arg here to `testPath` (singular of the array it's mapping)

jryanconklin:
@illepic Updated!

PR #210, opened 2017-12-29
Latest release badge to readme

illepic:


illepic:
Yarr, when this makes it to master, Greenkeeper will work.

PR #212, opened 2017-12-29
Jest testing with examples.

illepic:


.travis.yml
```diff
@@ -26,7 +26,7 @@ before_script:
 - composer --version
 - node --version
 - npm --version
-script: npm run lint && npm run compile:pl
+script: npm run lint && npm run test:unit && npm run compile:pl
```
jryanconklin: Nice! So we're including unit tests by default whereas pa11y is optional?

.eslintrc.js
```diff
@@ -1,7 +1,7 @@
 // Rule reference: http://eslint.org/docs/rules
 // Individual rule reference: http://eslint.org/docs/rules/NAME-OF-RULE
 module.exports = {
-  extends: "airbnb",
+  extends: ["airbnb", "plugin:jest/recommended"],
```
jryanconklin: It might be worth adding this to the comment above.

README.md
```diff
@@ -563,7 +563,7 @@ npm install --save-dev pa11y@5.0.0-beta.5 pa11y-reporter-cli
 Then whenever you want to run your tests, simply:
 
 ```bash
-npm run pa11y
+npm run test:accessibility
```
jryanconklin: Simply, `test:a11y` might be a nice shorthand here for folks.

source/_patterns/01-atoms/button/__tests__/button.test.js
```diff
@@ -0,0 +1,37 @@
+const $ = require('jquery');
```
jryanconklin: Tests are a part of our design system, and we don't expect these to move when / if demo folders and _twig-components move?

.travis.yml
illepic: pa11y current "fails". I'll get it included in Travis builds once it's passing.

source/_patterns/01-atoms/button/__tests__/button.test.js
illepic: Tests stay close to components. Bundles get deployed. Won't be affected by demo reshuffling.

PR #213, opened 2017-12-29
Bootstrap beta 3

illepic:
We shouldn't be including bootstrap libs at src, should be dist instead.

### DIFF
--- a/apps/pl/__tests__/pl.test.js
+++ b/apps/pl/__tests__/pl.test.js
@@ -1,0 +1,5 @@
+const BUILD_TARGET = 'pl';
+
+test('Pattern Lab requires global BUILD_TARGET of `pl`', () => {
+  expect(BUILD_TARGET).toBe('pl');
+});
--- a/jest.config.js
+++ b/jest.config.js
@@ -8,7 +8,9 @@ module.exports = {
     // Jest doesn't care about styles, images, fonts, etc
     '\\.(css|scss|jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
     '<rootDir>/tools/tests/unit/__mocks__/fileMock.js',
-    // Webpack aliases. @TODO: pull in webpack.shared.config.js and pull this piece off?
+
+    // Webpack aliases.
+    // @TODO: pull in webpack.shared.config.js and pull this piece off?
     '^protons$': '<rootDir>/source/_patterns/00-protons/',
     '^atoms$': '<rootDir>/source/_patterns/01-atoms/',
     '^molecules$': '<rootDir>/source/_patterns/02-molecules/',
--- a/jest.config.js
+++ b/jest.config.js
@@ -9,13 +9,13 @@ module.exports = {
     '\\.(css|scss|jpg|jpeg|png|gif|eot|otf|webp|svg|ttf|woff|woff2|mp4|webm|wav|mp3|m4a|aac|oga)$':
     '<rootDir>/tools/tests/unit/__mocks__/fileMock.js',
 
-    // Webpack aliases.
-    // @TODO: pull in webpack.shared.config.js and pull this piece off?
+    // Webpack aliases. This is a pain, but Babel does NOT understand Webpack resolved aliases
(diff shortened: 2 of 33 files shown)


illepic:
Blarg, moving to dist includes from bootstrap means there is no more parsing up the chain of node deps. Hrm, need a new approach here.

illepic:
Back to src includes. Babel is definitely happier.

apps/pl/__tests__/pl.test.js
```diff
@@ -0,0 +1,5 @@
+const BUILD_TARGET = 'pl';
```
grayside: I've never seen the __tests__ convention for a testing folder. Why need the underscores?

jest.config.js
```diff
@@ -8,12 +8,14 @@ module.exports = {
...
+
+    // Webpack aliases. This is a pain, but Babel does NOT understand Webpack resolved aliases
+    // @TODO: Look into helpers for this (e.g. https://github.com/mwolson/jest-webpack-alias)
     '^protons$': '<rootDir>/source/_patterns/00-protons/',
-    '^atoms$': '<rootDir>/source/_patterns/01-atoms/',
-    '^molecules$': '<rootDir>/source/_patterns/02-molecules/',
-    '^organisms$': '<rootDir>/source/_patterns/03-organisms/',
-    '^templates$': '<rootDir>/source/_patterns/04-templates/',
-    '^pages$': '<rootDir>/source/_patterns/05-pages/',
+    '^atoms[/](.+)': '<rootDir>/source/_patterns/01-atoms/$1',
```
grayside: Why define a character class for a single character?

package.json
```diff
@@ -24,7 +24,7 @@
     "compile:grav": "webpack --display-entrypoints --config webpack.grav.config.js",
     "lint:js": "eslint ./",
     "lint:scss": "stylelint '{source, apps/drupal, apps/pl}/**/*.scss'",
-    "lint": "npm run lint:js & npm run lint:scss",
+    "lint": "npm run lint:js && npm run lint:scss",
```
grayside: Is it on purpose that if JS linting fails it don't bother with SCSS linting? Alternatively could use a semicolon to run it in sequence, failing first only if the script it executed in a context where an error causes a script bail.

source/_patterns/00-protons/__tests__/protons.test.js
```diff
@@ -0,0 +1,7 @@
+import protons from '../';
+
+test('GLOBAL_CONSTANT of `blerp` is available', () => {
```
grayside: Since these test files are likely to be copied and used as the starting place for new test files, should there be more comments? Boring stuff like 

```
/**
 * These tests are focused on protons.
 *
 * The tests below are only a start, and should be expanded with X, Y, and Z for more completeness.
 */
```

apps/pl/__tests__/pl.test.js
illepic: Because Jest.

jest.config.js
illepic: My regex is not great. Are you asking about the `[/]`?

package.json
illepic: A quick note on this: Travis was exploding due to the simultaneous nature of the `&`. Moving to consecutive commands makes Travis happy, for some reason.

Regarding semicolon syntax, like this?

```sh
npm run lint:js;npm run lint:scss
```

source/_patterns/00-protons/__tests__/protons.test.js
illepic: I'll make a comment that `blerp` serves no other purpose than a starting point.

jest.config.js
grayside: Yes, I don't understand why it wouldn't just be this:

```
'^atoms/(.+)'
```

package.json
grayside: Si. Though conventionally there's a space after the semicolon.

jest.config.js
grayside: Note forward slash is a default regex delimiter in a lot of places, so that may have to be `^atoms\/(.+)` if babel gets complainy.

illepic: It works! Pushing updates now.

package.json
illepic: Moved to `;`

PR #214, opened 2018-01-02
Simple js widget example

illepic:
A simple example of how to build a test a widget that is mostly js.


illepic:
Needs tests.

source/_patterns/02-molecules/example-widget/index.js
```diff
@@ -0,0 +1,26 @@
+/**
+ * Example JS widget
+ *
+ * An example of a testable, standalone, javascript-driven widget that makes an api call, writes
+ * HTML, and does basic logic. This is a pretty naive re-render-on-data-change approach, using
+ * poor-man's Redux and jQuery
```
jryanconklin: I think this docblock could be expanded a little bit to explain future usages. I'm imagining the questions you'll be getting being best answered here. Specifically for this example widget it may be good to differentiate what we're doing in `src/index.js` vs `index.js`

source/_patterns/01-atoms/grid/_grid-3-up.twig
```diff
@@ -7,19 +7,19 @@
 
 {% if container or container_fluid %}<div class="{{ container_class }}">{% endif %}
   <div class="row">
-    <div class="col-sm-3">
+    <div class="col-sm-4">
```
jryanconklin: I for one will miss the more exciting columns. ⚰️

source/_patterns/02-molecules/example-widget/index.js
illepic: @jryanconklin are you looking for explanation around the idea of a source and an implementation?

jryanconklin: Yes indeed, I think even just saying "source and implementation" here would help as an example. Further since this will be the only example of this kind of widget, it might be good to list what the minimum requirements would be to make it work so it's replicable without people sending you basic questions.

illepic: @jryanconklin better docs here and at `example-widget/src/index.js`.

PR #221, opened 2018-01-17
Update pa11y to disable unwanted console output

jryanconklin:
This PR updates the console log to get rid of overflowing text. See https://github.com/pa11y/pa11y/issues/366

### DIFF
--- a/tools/tests/accessibility/pa11y.js
+++ b/tools/tests/accessibility/pa11y.js
@@ -21,11 +21,7 @@ const options = {
     'WCAG2AAA.Principle1.Guideline1_4.1_4_6.G18',
   ],
   log: {
-    debug: (message) => {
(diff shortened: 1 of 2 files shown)


tools/tests/accessibility/pa11y.js
```diff
@@ -21,7 +21,11 @@ const options = {
     'WCAG2AAA.Principle1.Guideline1_4.1_4_6.G18',
   ],
   log: {
-    debug: console.log,
+    debug: (message) => {
+      if (!message.startsWith('Browser Console:')) {
+        console.log(message);
+      }
```
grayside: Maybe we should have an in-line comment back to https://github.com/pa11y/pa11y/issues/366 given the comments by the maintainer about fixing related things, which may lead to a different solution in coming weeks/months.

PR #229, opened 2018-02-15
Update all dependencies to the absolute latest

illepic:


source/_data/scssVariables.json
```diff
@@ -1,7 +1,7 @@
 {
   "scssColors": {
     "primary": "rgb(0, 123, 255)",
-    "secondary": "rgb(134, 142, 150)",
+    "secondary": "rgb(108, 117, 125)",
```
jryanconklin: Interesting, I'm guessing this is coming over from the update to Bootstrap 4?

illepic: Bingo. Looks like Bootstrap 4, full release made a small tweak.

PR #233, opened 2018-02-18
Replace black color border for the bootsrap border color variable

yaazkal:
Borders here are black. If you define a different color for the $border-color varible it dosen't change because is hardcoded in the inline stye. This prevents the black color and uses whatever the color you define in the $border-color varible.


illepic:
@yaazkal ++ thank you! Merging.

yaazkal:
no problem, thanks to all of you !

PR #234, opened 2018-02-19
Adds class blockquote to the blockquote in the demo

yaazkal:
Acoording to the [Blockquote documentation](https://getbootstrap.com/docs/4.0/content/typography/#blockquotes) in Bootstrap  is necessary to add the blockquote class. This way it will shows correctly how the blockquote will look like.

Note that in the _type.scss you will edit `.blockquote { }`  and not   the html tag `blockquote`


illepic:
@yaazkal ++ thank you! Merging.

yaazkal:
Glad to contribute !

PR #244, opened 2018-02-27
Initial take on adding prettier formatting tool.

### LINKED ISSUE #132: Prettier.io auto-formatter for js, scss, twig?
[Webpack/gulp-task: prettier.io ](https://prettier.io/)

grayside:
Decided to poke around Particle and see if I could contribute something. Not sure if this really needs to be tied into webpack/gulp or not, here's a pure npm script implementation. The package.json and package-lock.json have been prettierified. Otherwise I left out the results of my testing, which tweaked a lot of files.

Fixes #132


grayside:
Prettier has it's own standard and [some options to tweak it](https://prettier.io/docs/en/options.html), but not nearly as comprehensive as eslint. I do not know if there's a package or blog post out there to help you align it with AirBnB.

The modifications to eslint config do two things:
1. Add some of prettier's formatting expectations to eslint so eslint can detect violations prettier would fix
2. Override existing eslint rules that conflict with prettier

My expectation would be that if you tweak prettier config to match behaviors of AirBnB you would need to replace the use of the extends configuration with cherrypicking from the prettier eslint rules.

illepic:
Well, hello: https://github.com/prettier/eslint-plugin-prettier

illepic:
@grayside , I've got this PR'd against the npm scripts updates from https://github.com/phase2/particle/pull/249 now.

Right off the bat, when in `npm run lint:js` I get this output:

image

So, it looks like Prettier now integrates directly to eslint which is awesome. The output is totally gnarly and hard to make out, but hopefully the `fmt` makes this a lot more fire-and-forget.

illepic:
Prettier'd all the things.

grayside:
👍

illepic:
Considering ignoring root package.json since every npm install rewrites the file. Also needs README.md entry about this.

illepic:
@grayside I added ignore for package and package-lock.

The last thing this needs is a bit of config to avoid this: Prettier wants to "line-pack" as much as possible. By that, I mean If we have an array of, say, 3 items stacked vertically:

```js
{
  'thing-1',
  'thing-2',
  'thing-3',
}
```

then Prettier wants to put all these to one line within the column count we have specified (100). This actually sucks for working in teams, since diffs get out of hand really quick. Other fmt's like Elm's encourage "tall" code specifically for the ease of working in version control. 

Can we configure Prettier to back off of this kind of optimization? Examples where I'd like to undo this are here:

- https://github.com/phase2/particle/pull/244/files#diff-b9e12334e9eafd8341a6107dd98510c9L9
- Note the specific `// prettier-ignore`: https://github.com/phase2/particle/pull/244/files#diff-b9e12334e9eafd8341a6107dd98510c9R113
- Note the specific `// prettier-ignore`: https://github.com/phase2/particle/pull/244/files#diff-b9e12334e9eafd8341a6107dd98510c9R128
- Note the specific `// prettier-ignore`: https://github.com/phase2/particle/pull/244/files#diff-b9e12334e9eafd8341a6107dd98510c9R142
- https://github.com/phase2/particle/pull/244/files#diff-9ad72506de2fbed30e32b56c56b4507aL7
- https://github.com/phase2/particle/pull/244/files#diff-d484380fb09d5fe07da65a4e76552cf6L11
- https://github.com/phase2/particle/pull/244/files#diff-d484380fb09d5fe07da65a4e76552cf6L11

illepic:
Also noting here that Prettier scss is grenading a few files that I need to look into.

illepic:
Keeping this alive. I've had a chance to use prettier on another project full time and I really like it.

source/_patterns/01-atoms/icon/scss/_icon.scss
```diff
@@ -8,23 +8,23 @@
...
 // Creates the font face tag for font icons.
 @font-face {
-  font-family: "iconfont";
+  font-family: 'iconfont';
   src: url('../font/iconfont.eot?cachebust=#{random(99999)}'); /* IE9 Compat Modes */
-  src: url('../font/iconfont.eot?#iefix') format('embedded-opentype'), /* IE6-IE8 */
-    url('../font/iconfont.woff') format('woff'), /* Pretty Modern Browsers */
-    url('../font/iconfont.ttf')  format('truetype'), /* Safari, Android, iOS */
-    url('../font/iconfont.svg') format('svg'); /* Legacy iOS */
+  src: url('../font/iconfont.eot?#iefix') format('embedded-opentype'),
```
tjheffner: all of the prettier formatting looks fine enough to me, this is the only section that gets annoying because the comments don't line up with the url & formats anymore. //prettier-ignore here?

illepic: Gah, good catch. These will be things we'll need to watch for, but it's worth it for the other 95% of code. I'll update now.

.huskyrc
```diff
@@ -0,0 +1,2 @@
+hooks:
+  pre-commit: "pretty-quick --staged"
```
tjheffner: so will this run prettier on every commit? :+1:

source/_patterns/01-atoms/icon/scss/_icon.scss
illepic: Actually, this gets nuked with @dslemay 's incoming icon work, so we'll let this slide.

.huskyrc
illepic: This does! I don't want prettier rewriting source files on webpack-dev-server, and this is a good "catch" to make sure it's run automatically at least once. Folks can still run `npm run fmt` manually to run it any time. Of course we need to document this ...

PR #251, opened 2018-03-04
pa11y v5 is full release, no need to depend on beta in readme

illepic:


illepic:
@grayside updated readme!

PR #257, opened 2018-03-07
Webpack twig loader, trigger PL build

illepic:
- Does NOT yet work for demo files (since they are not part of our design modules yet).


illepic:
Now works for all demo files, but blows up Jest tests. Researching.

tjheffner:
lot of new updates, recommend pulling & testing locally to check out the new article/homepage demo pages, and to poke around how the demo system is working (see `apps/pl/index.js`). the component generator should also generate a stub test file now.

PR #258, opened 2018-03-07
Task/grav webpack v4

jryanconklin:
This PR adds the file structure and commands created by @illepic for Drupal for our Grav version as a part of the update to Webpack 4.

Note: this still uses a manual step to update the twig-namespaces.yaml file in two undesirable ways:
* For use in Grav, this config file needs either to be copied or symlinked to `users/config/plugins/twig-namespaces.yaml`
* The `generated-namespaces` key must be updated manually to the right path (from `../..` to `user/themes/particle/`) and overwritten as the `namespaces` key. `generated-namespaces` are kicked off by gulp.


apps/grav/webpack.grav.shared.js
```diff
@@ -0,0 +1,26 @@
...
+const webpack = require('webpack');
+
+const grav = {
+  entry: {
+    'app-grav': [
+      path.resolve(__dirname, 'index.js'),
+    ],
+  },
+  // These will be explicitly be provided OUTSIDE the bundle via a <script> tag in the HTML
+  externals: {
```
tjheffner: Does grav include jquery & lodash by default, or are we doing that in our base grav templates already? I'm not seeing them actually included anywhere. otherwise i have no nits, everything looks good 👍

Issue #259, opened 2018-03-08
apps/drupal/particle.libraries.yml needs lots of work


illepic:
Currently apps/drupal/particle.libraries.yml looks like this:

```yml
# @TODO: THIS IS NEEDS LOTS OF WORK, THESE PATHS MAY BE NO LONGER CORRECT
core:
  css:
    theme:
      ../../dist/assets/app-drupal.styles.css:
        preprocess: true
  js:
    ../../dist/assets/commons.js:
      preprocess: true
    ../../dist/assets/app-drupal.js:
      preprocess: true
  # see all in Drupal's `core/core.libraries.yml`
  dependencies:
    # - patternlab/j-dash
    - core/drupal
# @TODO: GET THESE PULLED OUT VIA WEBPACK
#jquery:
#  js:
#    ../node_modules/jquery/dist/jquery.min.js: { minified: true, weight: -20 }
#lodash:
#  js:
#    ../node_modules/lodash/lodash.min.js: { minified: true, weight: -20 }
```

This needs to be revamped to:

1. Remove mention of commons.js
1. Use webpack to extract jquery and lodash to their own standalone files and referenced in the `dependencies` key to override core jquery and underscore.

Info: https://www.drupal.org/node/2497313

Shefarik:
I don't have assets/app-drupal.styles.css after compiling, I think we need have all styles from pattern there?

illepic:
@Shefarik on current beta.3 can you run `npm run build:drupal` and let me know what files you get in `dist/`.

jryanconklin:
@Bloom I think this issue can be closed?

PR #278, opened 2018-04-04
Task/grav theme updates

jryanconklin:
This PR updates the Grav theme included in Particle: 

* Renamed gravpl => particle for consistency
* Fix JavaScript console error
* Add Bootstrap NavBar Implementation


apps/grav/index.js
```diff
@@ -13,9 +13,12 @@ console.log('grav-theme ran');
 
 // Send each component the $(document) as its context
 const $context = $(document);
+const settings = {
+  enableHolder: TRUE,
+};
 
 // Let's just execute everything and pass in $(document), settings
 _.forEach(designSystem, (component) => {
```
illepic: @jryanconklin we've got a MUCH more elegant way to do this now, check the drupal/pl index.js files for deets (we now have api methods like `enableAllComponents` on the design-system library!).

apps/grav/templates/default.html.twig
```diff
@@ -1,18 +1,11 @@
...
   {% block content %}
 
     <div>
-      <p>Congratulations! This Grav theme is working! This is a solid proof of concept, but there's a ton more you can do here:</p>
-      <ul>
-        <li>Webpack the Grav</li>
-        <li>Update CSS Assets from Drupal -> Grav</li>
-        <li>Fix JavaScript</li>
-      </ul>
+      <p>Congratulations! This Grav theme is working!
```
illepic: Am I crazy or is there now closing `</p>`?

apps/grav/templates/partials/base.html.twig
```diff
@@ -24,13 +24,17 @@
...
+{% endblock %}
+
 <header class="site__header" role="banner">
   {% block header %}
     {% embed '@molecules/jumbotron/_jumbotron.twig' with { jumbotron_fluid: true } %}
       {% block jumbotron_branding %}
         {% include '@atoms/branding/_branding.twig' with {
-          site_name: 'Particle in ' ~ site.title|e('html'),
-          url: '/',
+          site_name: site.title|e('html'),
```
illepic: What is the `e()` twig filter here?

source/_patterns/02-molecules/nav/_nav.twig
```diff
@@ -32,11 +32,13 @@
...
 
 <{{ nav_element }} class="{{ nav_classes }}" {{ nav_id }} {{ nav_role }}>
-  {% for item in nav_items %}
-    {% if item.nav_item_dropdown %}
-      {% include '@atoms/dropdown/_dropdown.twig' with item %}
-    {% else %}
-      {% include '@molecules/nav/_nav-item.twig' with item %}
-    {% endif %}
-  {% endfor %}
+  {% block nav_items %}
```
illepic: Verifying that this refactor works back to pl/drupal apps?

apps/grav/templates/partials/base.html.twig
jryanconklin: I pulled this from the similar work done in Grav's antimatter theme. My guess was down to the fact that site.title could be set in headmatter on an admin menu. So since that value can accept user input, it should be escaped.

source/_patterns/02-molecules/nav/_nav.twig
jryanconklin: I spun up a Drupal env using this branch to check. Unfortunately I did it post merge of master. There is something wrong with the menu (the admin menu unsets on the page), but it looks like it's an issue with the underscore library as I'm getting a "Uncaught SyntaxError: Invalid or unexpected token" pointing to underscore in browser console.

I don't think it's the inclusion of the block is a problem for Drupal. Specifically because we had to do a bit of custom work to get this working in Drupal's menus such that we're actually not embedding or including `@molecules/nav/_nav.twig` in the Drupal theme. We write out what we need conditionally, then include the nav-items. I think this is okay, but we should open a ticket for underscore.

tl;dr I think this is still okay to merge.

apps/grav/templates/partials/base.html.twig
illepic: Ah, `e()` is escape!

source/_patterns/02-molecules/nav/_nav.twig
illepic: That underscore issue really should be solved here: https://github.com/phase2/particle/issues/259

illepic:
@jryanconklin where did the `apps/pl/index.js` changes go?

illepic:
Answered: was already in `master`.

PR #291, opened 2018-05-15
Particle issue #259 - apps/drupal/particle.libraries.yml needs lots of work

yuginaili:


apps/drupal/webpack.drupal.shared.js
```diff
@@ -7,15 +7,16 @@ const webpack = require('webpack');
...
+      path.resolve(__dirname, 'drupal-jquery.js'),
+    ],
+    'drupal-lodash': [
+      path.resolve(__dirname, 'drupal-lodash.js'),
+    ],
     'app-drupal': [
       path.resolve(__dirname, 'index.js'),
     ],
   },
-  // These will be explicitly be provided OUTSIDE the bundle via a <script> tag in the HTML
```
tjheffner: do we still need these as externals? or does our main bundle just find them from the other two entry points with no issues?

yuginaili: @tjheffner Correct, they should be made available to the main bundle by the two entry points with no hiccups

apps/drupal/drupal-lodash.js
```diff
@@ -0,0 +1,37 @@
...
+_.pluck = _.map;
+_.rest = _.drop;
+_.sample = _.sampleSize;
+_.select = _.filter;
+_.tail = _.drop;
+_.take = _.head;
+_.unique = _.uniq;
+_.where = _.filter;
+
+window._ = _;
```
illepic: Why `window` here, but `global` in `drupal-jquery.js`?

apps/drupal/drupal-lodash.js
```diff
@@ -0,0 +1,37 @@
+import _ from 'lodash';
+
+// Convert Underscore methods to Lodash.
```
illepic: Bravo. 👏

apps/drupal/particle.libraries.yml
```diff
@@ -1,22 +1,24 @@
...
-#    ../node_modules/jquery/dist/jquery.min.js: { minified: true, weight: -20 }
-#lodash:
-#  js:
-#    ../node_modules/lodash/lodash.min.js: { minified: true, weight: -20 }
+    - core/drupalSettings
+
+# Create a custom jQuery and Lodash libraries
+# that will override the Drupal core jQuery and Underscore
+# See `libraries-override` in particle.info.yml
+particle-jquery:
```
illepic: A suggestion, now that we have this rock solid, using `particle/particle-jquery|lodash` in `particle.info.yml` now feels weird. Let's get the library updated so we can just do `particle/jquery` and `particle/lodash` now.

Basically, just update the libraries key, no need to change the webpack output filenames ;)

apps/drupal/webpack.drupal.dev.js
```diff
@@ -24,7 +24,7 @@ const dev = {
   plugins: [
     new WebpackShellPlugin({
       onBuildEnd: [
-        'echo \nWebpack drupal dev build complete! Edit apps/drupal/webpack.drupal.dev.js to replace this line with `drupal cr all` now.',
+        'drupal cr all',
```
illepic: Let's change this so that we have the echo, but also the `drupal cr all` commented out. Make a note that a user will need to uncomment it to use it. (We can't assume everyone will have drupal console installed.

apps/drupal/webpack.drupal.shared.js
illepic: I'm going to pull this down and verify.

illepic:
Like @jryanconklin said: Between the info and libraries file, let's explain the method to our madness here. It's super powerful, but we definitely need to explain it.

apps/drupal/webpack.drupal.dev.js
grayside: I'm not sure what the trends are, but my expectation is still that Drush is more common.

apps/drupal/particle.libraries.yml
```diff
@@ -1,22 +1,24 @@
...
-# @TODO: GET THESE PULLED OUT VIA WEBPACK
-#jquery:
-#  js:
-#    ../node_modules/jquery/dist/jquery.min.js: { minified: true, weight: -20 }
-#lodash:
-#  js:
-#    ../node_modules/lodash/lodash.min.js: { minified: true, weight: -20 }
+    - core/drupalSettings
+
+# Create a custom jQuery and Lodash libraries
```
illepic: >Create custom

(No "a")

apps/drupal/particle.libraries.yml
```diff
@@ -1,22 +1,24 @@
...
+# that will override the Drupal core jQuery and Underscore
+# See `libraries-override` in particle.info.yml
```
illepic: Perfect.

apps/drupal/webpack.drupal.dev.js
```diff
@@ -24,7 +24,7 @@ const dev = {
   plugins: [
     new WebpackShellPlugin({
       onBuildEnd: [
-        'echo \nWebpack drupal dev build complete! Edit apps/drupal/webpack.drupal.dev.js to replace this line with `drupal cr all` now.',
```
illepic: We have to leave this line in

apps/drupal/particle.info.yml
```diff
@@ -5,6 +5,9 @@ base theme: stable
 core: 8.x
 libraries:
   - particle/core
+libraries-override:
```
illepic: Need a note here about what we're doing and the potential ramifications (like jQuery compatibility, lodash/underscore compatibility).

PR #299, opened 2018-06-02
Sass import fixes, upgrades to packages.

illepic:


jryanconklin:
So the fix here is adding css in test order to compile in the expected order?

PR #302, opened 2018-06-06
Relative pathing url added to demonstration bg images.

illepic:
Relative static asset pathing for things like css background images requires new webpack loader and a bit of ExtractTextPlugin config


source/_patterns/01-atoms/image/demo/image-styles.twig
```diff
@@ -1,15 +1,16 @@
 {% extends '@atoms/grid/_grid-1-up.twig' %}
```
illepic: Noting that all changes in this file are purely indentation :P

PR #306, opened 2018-06-07
Adding resolve-url-loader to Prod config

yuginaili:


illepic:
Just verifying: if this is left out, `build` fails right?

PR #308, opened 2018-06-09
[WIP] Convert run-script-on-filetype-change to new Webpack 4 plugin api.

illepic:
Work in progress. No need to review yet.


illepic:
Noting that we need some big updates to the official docs. Leaning heavily on `npx` now to run gulp tasks.

Issue #310, opened 2018-06-12
Rip out lodash


illepic:
We should remove lodash.

- It is barely used throughout Particle
- It becomes a library to juggle against consuming apps (lookin' at you, Drupal)
- There are ES6 equivalent features anyway

illepic:
We've relegated lodash to only devDependencies for use in yeoman and other tools. Those tools will themselves become individual packages eventually, so we'll be removing many devDependencies soon.

PR #320, opened 2018-06-18
Migrate lodash to devdependencies

dslemay:
https://github.com/phase2/particle/issues/310

Some of the usage of lodash has been replaced with Vanilla JS. The remaining methods of lodash are all within `tools/tasks/gulp-twig-namespaces.js` and offer a more succint means of completing the functionality. Lodash has also been migrated to devDependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,8 +43,6 @@
     "font-awesome": "^4.7.0",
     "holderjs": "^2.9.4",
     "jquery": "^3.3.1",
-    "lodash": "^4.17.10",
-    "lodash.camelcase": "^4.3.0",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -76,6 +73,8 @@
     "iconfont-plugin-webpack": "^1.1.0",
     "jest": "^22.4.3",
     "js-yaml": "^3.11.0",
+    "lodash": "^4.17.10",
+    "lodash.camelcase": "^4.3.0",
```
illepic: We don't need camelCase if we're including all of lodash for devDeps.

PR #358, opened 2018-09-05
Move the globbing loader to pl specifically.

illepic:
Handles: https://github.com/phase2/particle/issues/357


particle.js
```diff
@@ -31,8 +27,8 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 // - development
 // - production
 const { NODE_ENV } = process.env;
-// Paths
-const { PATH_SOURCE } = require('./config');
+const { PATH_PATTERNS, PATH_DIST } = require('./config');
+// console.log(PATH_PATTERNS);
```
dslemay: Stray console statement

gulpfile.js
```diff
@@ -33,56 +34,68 @@ gulp.task('compile:pl', plCompile);
  */
 const twigNamespaces = require('./tools/tasks/gulp-twig-namespaces');
 
+/**
+ * Gulp needs to work in relative paths, NOT absolute like we've configured in
```
jryanconklin: Grav needs an absolute path for its generated spaces. We're currently using a workaround to achieve what we need for the `twig-namespaces` plugin. See `apps/grav/twig-namespaces.yml` for the paths we need (under `namespaces` key). Eventually the path for this would need to be something like: `user/themes/particle/source/_patterns/01-atoms`

illepic: Aaaand, solved :) Check out the changes to gulp-twig-namespaces.js and gulpfile.js

PR #366, opened 2018-09-14
Update all npm packages, upgrade jest to 23.

illepic:


package.json
```diff
@@ -71,17 +71,17 @@
...
-    "lodash": "^4.17.10",
+    "lodash": "^4.17.11",
     "mini-css-extract-plugin": "^0.4.2",
     "node-sass": "^4.9.3",
     "postcss-loader": "^3.0.0",
     "prettier": "^1.14.2",
     "pretty-quick": "^1.6.0",
     "prismjs": "^1.15.0",
-    "resolve-url-loader": "^2.3.0",
+    "resolve-url-loader": "^3.0.0",
```
jryanconklin: Consider for 3.0 here (https://www.npmjs.com/package/resolve-url-loader):
```
Breaking Changes

    Multiple options changed or deprecated.
    Removed file search "magic" in favour of join option.
    Errors always fail and are no longer swallowed.
    Processing absolute asset paths requires root option to be set.
```

illepic: All set.

PR #387, opened 2018-09-26
Add vue-testing-library and testing for vue-widget

Lawlheart:
Adds vue-testing-library and some basic tests to the example vue widget


jest.config.js
```diff
@@ -13,10 +13,15 @@ module.exports = {
     // Protons use a slightly different pattern because it's the only source pattern directory that
     // has a base-level index.js (this is how all components ensure access to base-level styles & variables).
     '^protons$': '<rootDir>/source/_patterns/00-protons/',
```
illepic: Can we just remove this line?

package.json
```diff
@@ -59,6 +59,7 @@
     "babel-core": "^7.0.0-bridge.0",
     "babel-jest": "^23.6.0",
     "babel-loader": "^8.0.2",
+    "babel-polyfill": "^6.26.0",
```
illepic: Why is this being added if we already had it before (see `particle.js` prepending it to entry points).

tools/tests/unit/setupJest.js
```diff
@@ -1,2 +1,5 @@
+// TODO: see if we can make this import conditional based on node version
```
illepic: Nuke your comment here, I don't think that's what you mean.

jest.config.js
LawlietBlack: I think so, but I wasn't sure. Probably fine if you think we can

package.json
LawlietBlack: I think we replaced it with @babel/preset-env

tools/tests/unit/setupJest.js
LawlietBlack: sure thing

jest.config.js
```diff
@@ -13,10 +13,15 @@ module.exports = {
...
     '^atoms/(.+)': '<rootDir>/source/_patterns/01-atoms/$1',
     '^molecules/(.+)': '<rootDir>/source/_patterns/02-molecules/$1',
     '^organisms/(.+)': '<rootDir>/source/_patterns/03-organisms/$1',
     '^templates/(.+)': '<rootDir>/source/_patterns/04-templates/$1',
     '^pages/(.+)': '<rootDir>/source/_patterns/05-pages/$1',
   },
+  transform: {
+    '^.+\\.js$': '<rootDir>/node_modules/babel-jest',
+    '.*\\.(vue)$': '<rootDir>/node_modules/jest-vue-preprocessor',
+  },
```
LawlietBlack: Double check this? I'm not 100% sure I did this part right.

source/_patterns/02-molecules/vue-widget/src/vue-cryptos/components/facet-table-facets.vue
```diff
@@ -36,7 +36,6 @@ export default {
   methods: {
     setFilter(facet) {
       this.$emit('updateFilter', facet);
-      this.filter = facet;
```
LawlietBlack: Removed this line because the parent sets the facet with props.

illepic:
Pull in master to kill that package.json conflict.

LawlietBlack:
Ready for review

package.json
dslemay: This should be removed since it is for version 6 of Babel and we are on version 7. @babel/env on line 60 at this PR is the package we need. It is being prepended in [particle.js](https://github.com/phase2/particle/blob/master/particle.js#L80).

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
```diff
@@ -0,0 +1,85 @@
...
+    };
+
+    const { getByText } = render(banner, { props });
+    expect(getByText('krieger', { exact: false }).textContent).toContain(
+      'krieger'
+    );
+  });
+
+  it('changes color on click', () => {
+    const bannerElement = document.querySelector('.vue-banner');
```
dslemay: Curious why this test doesn't use the vue-testing-library. Also, [jest-dom](https://github.com/gnapse/jest-dom) can import some nice selectors that may also be good to include in the project. Larger point of conversation that may expand past this PR.

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-clock.spec.js
```diff
@@ -0,0 +1,33 @@
...
+    const { getByTestId } = render(vueClock);
+    const hoursHand = getByTestId('hours-hand');
+    expect(hoursHand.style.transform).toContain('rotate');
+  });
+
+  it('updates the seconds hand', done => {
+    const { getByTestId } = render(vueClock);
+    const secondsHand = getByTestId('seconds-hand');
+    const before = secondsHand.style.transform;
+    setTimeout(() => {
```
dslemay: I think this probably is a good use case for using [Jest's fake timers](https://jestjs.io/docs/en/timer-mocks.html) to handle these assertions since we can manually advance the timer. Happy to help take a look at it as well. This should also eliminate the need for the done callback as well.

tools/tests/unit/setupJest.js
```diff
@@ -1,2 +1,4 @@
+import 'babel-polyfill';
```
dslemay: This should import @babel/polyfill since we are on Babel v7.

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
LawlietBlack: I don't think I knew how to target classes with it, and I think at the time I was trying to avoid adding testid. I can change it to testid though

LawlietBlack: As for babel, is there a way to import it like I'm doing in the jest config? I know it's a bit redundant, but I wasn't sure if there was another way

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-clock.spec.js
LawlietBlack: Ah, yeah, that would be great, didn't know that was a thing!

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
LawlietBlack: Oh, I see that we already had @babel/polyfill now. D'oh!

dslemay: Yep. We just need to remove the v6 version of the polyfill from the dependencies and import the babel namespaced import that is for v7.

dslemay: Benefit I could see of using testids is that it also maintains a consistent testing approach. There's not anything wrong with the approach, just my thoughts as I have been heavily steeped in Jest testing lately.

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-clock.spec.js
LawlietBlack: So, after meeting with Daniel, it looks like because the element uses Date.now to keep track of time, the jest timing mocks don't work on it. Going to keep this as is for now.

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
LawlietBlack: So, troubleshooting some odd behavior here - testid doesn't seem to be working on just this test: 

```
banner.vue › changes color on click

    Unable to find an element by: [data-testid="vue-banner"]

    <div
      class="vue-banner"
      data-testid="vue-banner"
      style="background: rgb(0, 0, 0);"
    >
      <h3>
        <marquee>
          <span>
            krieger was clicked!
          </span>
        </marquee>
      </h3>
    </div>

```

package.json
LawlietBlack: done :-)

jest.config.js
LawlietBlack: Okay, no, we can't just remove it. 
`Cannot find module 'protons' from 'index.js'`
maybe a different regex pattern?

illepic: Where does this error happen? I want a single "protons" in this list, not 2.

package.json
illepic: Why is this still in?

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
```diff
@@ -0,0 +1,85 @@
+import { render, Simulate } from 'vue-testing-library';
```
illepic: Is `integration/` necessary? Can it be removed to drop a little complexity?

LawlietBlack: yeah, sure thing

jest.config.js
LawlietBlack: removing that line makes 18 tests fail, so it's in a lot of places

LawlietBlack: I'm experimenting with regex patterns to match both cases. If I'm right, we just need a regex pattern that matches these, right?:

* protons
* protons/
* protons/<thing>

tjheffner: i.e. can we remove line 16, not line 15. the comment in line 13/14 explains why line 15 is needed

LawlietBlack: This line should do it, I think? Double check @dslemay , @illepic :
`'^protons[/]?(.*)': '<rootDir>/source/_patterns/00-protons/$1',`

source/_patterns/02-molecules/vue-widget/__tests__/integration/vue-cards.spec.js
LawlietBlack: I'm rolling this back to the previous version for now

jest.config.js
LawlietBlack: Line 16 was needed to match protons/thing => protons/thing, specifically for the utilities.js file that is there. I think this regex pattern should match all our needs

PR #402, opened 2018-10-04
require.context() tune ups for better performance.

illepic:


tools/webpack/import-all.js
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * results in:
+ *
+ *   components['vue-widget'] = {name: 'vue-widget', enable() {}, disable() {}}
+ *
+ * @param context
+ */
+function importAll(context) {
+  return context.keys().reduce((accumulator, componentPath) => {
+    // console.log(componentPath);
```
dslemay: Stray console log

tools/webpack/import-all.js
```diff
@@ -0,0 +1,26 @@
...
+    // "require" the component
+    const component = context(componentPath);
+    // Add a key to the components object that is the component's name, and a
+    // value that is full component
+    // componentsAccumulator[component.name] = component;
+    return Object.assign(accumulator, { [component.name]: component });
```
dslemay: One question/concern that I have is if we should flag that this assumes that each component has a unique name, else the latest will override a previous key/value pair. I recognize I may be overthinking the use case here, but with the magic added maybe a comment to that effect would be helpful?

tools/webpack/import-all.js
illepic: I'm leaving this as a key reference point to debugging.

tools/webpack/import-all.js
illepic: I'll make a note that this presumes unique component names. The SUPER magic solution is that is concatenates component type + component into the key ...

tools/webpack/import-all.js
dslemay: 👍

tools/webpack/import-all.js
jryanconklin: I'm curious here how we should resolve in the meantime several modifiers that have their own unique components. Say, for example, someone was converting a venerable code base to Particle.

illepic: @jryanconklin ensure each of your components has an `index.js` file in the format of current Particle components (run `npm run new` to get a feel for the basic requirements).

Then in `design-system.js` I'd say manually import each of your components with 

```js
import * as componentName from 'atoms/componentName';
// ...
export default {
  componentName,
};
```

PR #410, opened 2018-10-05
Updated link in `welcome.twig` to open in a new tab

coraunderhill:
I've set the link to the Github page on the welcome page to open in a new tab. Previously the link was trying to open within the `iframe` causing the page to stop operating.


illepic:
Noting here: use branch names like `feature/feature-name`.

PR #411, opened 2018-10-07
Big bootstrap refactor, other QoL fixes

illepic:
There are a few goals here.

###  Moving anything having to do with Bootstrap up to `protons`. 

Instead of spreading `@imports` and Bootstrap config overrides all over the design system, there are now exactly two files you modify:

1. *non-printing/_bootstrap-overrides.scss*: https://github.com/phase2/particle/pull/411/files#diff-1d4dfe8e79efd89926f706ae8ba99ecd
1. *printing/_bootstrap-printing.scss*: https://github.com/phase2/particle/pull/411/files#diff-99fc98263502a7e6fb55475791ae268f

We now do all variable overrides in ***_bootstrap-overrides.scss*** and actual printing inclusion in ***_bootstrap-printing.scss***.  This is a much more sustainable model in that it means ripping out Bootstrap is easier and bringing over other libraries like, say, Tailwind are much more straightforward.

Developers are still free to move individual component imports to their respective Particle components, if they so desire. For example, one could move `@import '~bootstrap/scss/card';` from within *_bootstrap-printing.scss* to *02-molecules/card/_card.scss*.

### Custom font example

Particle now features the Oswald font for headings via a simple non-printing Bootstrap variable override in *non-printing/_bootstrap-overrides.scss* (https://github.com/phase2/particle/pull/411/files#diff-1d4dfe8e79efd89926f706ae8ba99ecdR11) and an actual printing `@import` inclusion in *printing/_type.scss* (https://github.com/phase2/particle/pull/411/files#diff-7ae044a706f7dc5b92702edbf7565514R7).

Please note that *printing/_type.scss* features **BOTH** remote and local examples for including fonts.

### Webpack and gulp are further separated

The modifications to *webpack.pl.js* (https://github.com/phase2/particle/pull/411/files#diff-063f3480d0f3215a010ba32ab2b61b44) mean we no longer trigger gulp PL builds from within a webpack file.

### Root index.html

Visiting http://localhost:8080/ will no longer throw an error, and will instead offer you a link to `/pl`.


jryanconklin:
This is a good change.

illepic:
@jryanconklin, got it started here: https://phase2.gitbook.io/frontend/particle/sprint-0

source/_patterns/00-protons/non-printing/_colors.scss
```diff
@@ -1,9 +1,4 @@
...
 
-// Bootstrap overrides
-// Add colors to $theme-colors var here. Find other vars to override at ~bootstrap/scss/_variables.scss
-// Be sure to add new colors in quotes.
-$theme-colors: (
-  'test': pink,
-  'new': green,
-);
+$color-blerp: pink;
+$color-derp: green;
```
tjheffner: this implies we need to add colors vars in two places now--- initially as a variable here, and then referenced again in the _bootstrap-overrides to get all the extra goodies that bootstrap does with the $theme-colors map (e.g. generate css vars, color helper classes). i'm not opposed but that could get annoying / provide another point of failure if people aren't aware

source/_patterns/00-protons/printing/_bootstrap-printing.scss
```diff
@@ -0,0 +1,45 @@
...
+
+//@import "~bootstrap/scss/functions"; //_bootstrap-overrides.scss
+//@import "~bootstrap/scss/variables"; //_bootstrap-overrides.scss
+//@import "~bootstrap/scss/mixins"; //_bootstrap-overrides.scss
+@import '~bootstrap/scss/root';
+@import '~bootstrap/scss/reboot';
+@import '~bootstrap/scss/type';
+@import '~bootstrap/scss/images';
+@import '~bootstrap/scss/code';
+@import '~bootstrap/scss/grid';
```
tjheffner: i see the note about including everything here for simplicity, but it seems like a solid vote for moving grids from atoms to protons as a "printing" config or demo

source/_patterns/01-atoms/grid/_grid.scss
```diff
@@ -1,10 +1,10 @@
 // The non-printing variables, functions, and mixins of 00-protons/variables
 // are provided to this file via:
-// particle.js: sass-loader.data|includePaths
+//   webpack.particle.js: sass-loader.data|includePaths
 
 // The non-printing variables, functions, and mixins of 00-protons/variables
 // are provided to this file via:
-// particle.js: sass-loader.data|includePaths
+//   webpack.particle.js: sass-loader.data|includePaths
```
tjheffner: this file has duplicated comment

tjheffner: and still imports grid below

source/_patterns/00-protons/non-printing/_colors.scss
illepic: In a way I prefer this for this reason:

Our custom colors are our own, totally independent of Bootstrap. We then give them to Bootstrap specifically.

This slowly lets us move down the path to renaming `bootstrap-printing` and `bootstrap-overrides` to `vendor-printing` and `vendor-overrides`.

source/_patterns/00-protons/printing/_bootstrap-printing.scss
illepic: Our grids actually have usable twig patterns which are used a ton. There is no includable twig that lives in protons.

tjheffner: that is true, but it feels a little weird to me still that most of our protons demos rely on the grid atom (or the card sass, for colors). its a semantic point, but a grid system seems like part of project config. not a blocker here

illepic: Ah, good point: a proton demo that needs an atom ... Hrm...

illepic: Here are my stated reasons for keeping grid an atom:

1. Protons are variable/styling/logical generating **only**. The *first* concept in our take on Atomic Design that actually `{% include|embed|extends %}`'s are atoms. 
1. Note the structure of protons vs the other categories: 
image. Protons includes no other standalone components, on purpose.
1. Protons only uses grid *as part of its demos*.  Grid is not being used as an actual inclusion in another component, just for a demo purpose. We should not conflate inclusion in a demo for inclusion in a component!

jryanconklin: This reasoning makes sense to me and I'm good with it! 

May I offer as an alternative:
1) Remove use of the grid pattern in the proton demos and leave un-styled or (preferably);
2) Write Out default bootstrap in proton demos, e.g., manually write out `<div class="container">`. 

My reason being is that I feel we should be able to reliably delete components top-down without any breakages or webpack errors. e.g, Delete components from pages => atoms without the user having to diagnose what broke their build.

illepic: I don't hate this.

jryanconklin: If you like, I can open a separate issue for this this and tackle it after this current Bootstrap PR is merged?

illepic: Naw, I'll get this in on this PR. This will be good code to write today while I should be paying attention in meetings.

illepic: Done. https://github.com/phase2/particle/pull/411/commits/d69162e375393c146c1223c667a48a70375419b0

source/_patterns/00-protons/non-printing/_bootstrap-overrides.scss
```diff
@@ -0,0 +1,22 @@
...
+// Buttons
+$btn-border-radius: 0.25rem;
+
+// Cards
+$card-border-width: 1px;
+
+// Bootstrap non-printing tools. DO NOT REMOVE.
+@import '~bootstrap/scss/functions';
+@import '~bootstrap/scss/variables';
+@import '~bootstrap/scss/mixins';
```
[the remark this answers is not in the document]
illepic: @josemanosalvas, good catch. This broken out structure lets us do it the way you've mentioned. I'll update this stat.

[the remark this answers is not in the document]
illepic: https://github.com/phase2/particle/issues/420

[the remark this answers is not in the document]
illepic: Ah, yes, I remember now why it's in this order:

Bootstrap **really** wants variable overrides to come before Bootstrap imports for this reason:

image

Since Bootstrap does processing on overridden vars the overrides have to come before. HOWEVER, this might be a workaround:

image

This gives us the initial vars available to our overrides and then allows us to override. It's not **too** gross. What do you say @josemanosalvas ?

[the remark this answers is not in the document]
josemanosalvas: Seems like a good solution to me!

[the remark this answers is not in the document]
illepic: @josemanosalvas so, bad news ...

Doing this kind of double-`_variables.scss` import seems to cause a divergence in Bootstrap. receiving different values for `theme-color('primary');` if there exists this twice-imported `_variables.scss`.

Still investigating but leaning to only importing AFTER overrides (what is already in this PR).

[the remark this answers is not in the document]
Jwaxo: @illepic, @josemanosalvas, my solution so far has to been to just move the `_functions` and `_mixins` imports below the bootstrap override. Still doesn't quite fix `_colors`, `_spacing`, and `_type`, but there's really only so much you can do to pull these things out.

PR #412, opened 2018-10-07
[WIP] Fontawesome 5

illepic:


source/_patterns/01-atoms/svgicon/fontawesome.js
```diff
@@ -0,0 +1,21 @@
...
+ * https://fontawesome.com/how-to-use/with-the-api/setup/getting-started
+ */
+
+import { library, dom } from '@fortawesome/fontawesome-svg-core';
+// Import specific icons required
+import {
+  faUserAstronaut,
+  faRocket,
+  faSpaceShuttle,
+  faUser,
```
tjheffner: i like the theme here with spacegoat, but we should note these need to be changed in the demo also if they do get removed. or include more-standard ones by default, that projects are likelier to use (search, bars, times, user)

PR #413, opened 2018-10-08
cssnano squishes css on production builds

illepic:


webpack.particle.js
```diff
@@ -56,27 +57,37 @@ module.exports = {
         use: [
           {
             loader: 'css-loader',
-            options: { sourceMap: true, importLoaders: 2 },
```
dslemay: Why are we removing the importLoaders key?

illepic: It's not actually needed! This was once required in order to get source maps to transfer loaders that didn't have source maps. Now that url-loader properly provides source maps, we can ditch this.

PR #423, opened 2018-10-10
Add Rem-Calc to non-printing/_functions

Jwaxo:
This is a helpful function that uses MATH to convert all px and other measurements to rems. It can optionally be set via a variable; I defaulted it to 16px here.


source/_patterns/04-templates/basic-page/_basic-page.scss
```diff
@@ -10,18 +10,18 @@
...
+    padding: rem-calc(50px 20px);
 
     @include media-breakpoint-up(md) {
-      padding: 50px;
+      padding: rem-calc(50px);
     }
 
     p {
-      line-height: 2;
+      line-height: rem-calc(32px);
```
illepic: The `line-height` value without units is a multiplier of the font-size, so no need to convert to rem.

Jwaxo: @illepic Goddangit this is some knowledge that dropped to the recesses of my brain. Time to go rewrite all of my stuff.

source/_patterns/00-protons/non-printing/_functions.scss
```diff
@@ -1 +1,31 @@
...
+
+/**
+ * A "use everywhere" function that converts any standard CSS measurement to rem.
+ *
+ * Call with 1, 2, 3 or 4 parameters. 'px' is not required but supported
+ * @example rem-calc(10 20 30px 40);
+ * Numbers are space delimited. If you want to delimit using commas, wrap it in another pair of parens.
+ * @example rem-calc((10, 20, 30, 40px));
+ */
+@function rem-calc($values, $base-value: 16px) {
```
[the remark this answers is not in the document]
Jwaxo: @illepic perfect! I looked everywhere for one defined, but for some reason didn't assume Bootstrap has one.

source/default/_data/scssVariables.json
```diff
@@ -19,9 +19,9 @@
     "xl": "1200px"
   },
   "scssSpacing": {
-    "s": "5px",
-    "m": "25px",
-    "l": "50px"
+    "s": "rem-calc(5px)",
+    "m": "rem-calc(25px)",
+    "l": "rem-calc(50px)"
```
illepic: Blaaaarg, the sass2json library is exploding this. We need to toss out these spacing Sass vars anyway (there's already a mixin within Bootstrap providing exactly this that uses all the proper rem values).

I can help remove this sass spacing atom.

Issue #426, opened 2018-10-11
Signal Drupal App's Dependency on Component Libraries Module 


jryanconklin:
### Task Description

Integration with Drupal requires the [Component Libraries](https://www.drupal.org/project/components) module to work as expected. On installation the theme should warn the user if Components is not installed.

### Acceptance Criteria

On a Particle theme installation in Drupal 8, a message is displayed on the screen if the Component Libraries module is not installed.

#### References

https://drupal.stackexchange.com/questions/188322/how-can-i-check-if-a-module-is-enabled
https://www.drupal.org/node/2937955

PR #427, opened 2018-10-12
Components Error Handling and Page

jryanconklin:
See Issue https://github.com/phase2/particle/issues/426

This PR adds a generic template variable that checks whether or not the Component Libraries module is installed. If not, the page is redirected to a more "helpful" error page.

### DIFF
--- a/apps/drupal/particle.theme
+++ b/apps/drupal/particle.theme
@@ -13,7 +13,7 @@ use Drupal\Core\Form\FormStateInterface;
 function particle_preprocess_page(&$variables) {
   $moduleHandler = \Drupal::service('module_handler');
   if ($moduleHandler->moduleExists('components')){
-    $variables['components'] = TRUE;
+    $variables['has_components_module'] = TRUE;
   }
 }
 
--- a/apps/drupal/particle.theme
+++ b/apps/drupal/particle.theme
@@ -8,6 +8,16 @@
 use Drupal\Core\Form\FormStateInterface;
 
 /**
+ * Implements hook_preprocess_HOOK().
+ */
+function particle_preprocess_page(&$variables) {
+  $moduleHandler = \Drupal::service('module_handler');
+  if ($moduleHandler->moduleExists('components')){
+    $variables['components'] = TRUE;
+  }
+}
+
+/**
  * Implements hook_theme_suggestions_HOOK_alter().
  */
 function particle_theme_suggestions_block_alter(array &$suggestions, array $variables) {
--- a/apps/drupal/templates/layout/components--error-message.html.twig
+++ b/apps/drupal/templates/layout/components--error-message.html.twig
@@ -41,7 +41,7 @@ body {
 </style>
 
 <div class="components_error">
-  <h1 class="components_error--blinker">CRITIAL FAILURE OMG</h1>
+  <h1 class="components_error--blinker">CRITICAL FAILURE OMG</h1>
   <p>It looks like you forgot to install the  <a href="https://www.drupal.org/project/components">Component Libraries</a> module. Please install it and come back! Have a great day!</p>
 </div>
 <img class="components_error--image" src="https://media1.giphy.com/media/JIX9t2j0ZTN9S/giphy.gif?cid=3640f6095bbff9f35873486a55d52ada" alt="funny cat GIF">
--- a/apps/drupal/templates/layout/components--error-message.html.twig
+++ b/apps/drupal/templates/layout/components--error-message.html.twig
@@ -1,0 +1,47 @@
+{#
+/**
+ * @file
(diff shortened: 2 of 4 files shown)


apps/drupal/particle.theme
```diff
@@ -7,6 +7,16 @@
 
 use Drupal\Core\Form\FormStateInterface;
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function particle_preprocess_page(&$variables) {
+  $moduleHandler = \Drupal::service('module_handler');
+  if ($moduleHandler->moduleExists('components')){
+    $variables['components'] = TRUE;
```
illepic: Oh man, a root page var named `components` is going to collide *real* quick. How about `has_components_module` or something Drupal standardized equivalent?

apps/drupal/templates/layout/components--error-message.html.twig
```diff
@@ -0,0 +1,47 @@
...
+    opacity: 0;
+  }
+}
+</style>
+
+<div class="components_error">
+  <h1 class="components_error--blinker">CRITICAL FAILURE OMG</h1>
+  <p>It looks like you forgot to install the  <a href="https://www.drupal.org/project/components">Component Libraries</a> module. Please install it and come back! Have a great day!</p>
+</div>
+<img class="components_error--image" src="https://media1.giphy.com/media/JIX9t2j0ZTN9S/giphy.gif?cid=3640f6095bbff9f35873486a55d52ada" alt="funny cat GIF">
```
illepic: 😍

apps/drupal/templates/layout/page.html.twig
```diff
@@ -9,6 +9,7 @@
  * Available variables:
  *
  * General utility variables:
+ * - components: The boolean value of the components module requirement.
```
illepic: And here.

apps/drupal/templates/layout/components--error-message.html.twig
```diff
@@ -0,0 +1,47 @@
...
+ * @file
+ * Theme to display an error page for missing the Component Libraries module.
+ *
+ * @see template_preprocess_page()
+ * @see html.html.twig
+ */
+#}
+
+<style>
+@import url('https://fonts.googleapis.com/css?family=Roboto+Condensed');
```
illepic: Does this not inherit the rest of the design system's CSS?

apps/drupal/templates/layout/page.html.twig
```diff
@@ -44,6 +45,9 @@
  */
 #}
 
+{% if not components %}
+  {% include '@particle/layout/components--error-message.html.twig' %}
```
illepic: This `@particle` namespace: is this Drupal core providing the path to the theme?

tjheffner: yes, this is a drupal standard

apps/drupal/templates/layout/components--error-message.html.twig
tjheffner: possibly move this to drupal-styles.scss

jryanconklin: The reason I put the styles inside was to make sure this template is very easy to throw away. Plus, this is likely the first thing a user is going to see on installing the theme. There's no guarantee at this step that they've webpacked the CSS assets for drupal. 

Hopefully someone only ever sees this once, installs the module and deletes this file.

illepic: Ah, in that case go with a good ol' system font stack:

`-apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji" !default;`

apps/drupal/templates/layout/components--error-message.html.twig
yuginaili: Pretty nitpicky - would it be preferable to put this file within templates/misc

apps/drupal/templates/layout/components--error-message.html.twig
jryanconklin: Done!

apps/drupal/templates/misc/components--error-message.html.twig
```diff
@@ -0,0 +1,47 @@
...
+
+@keyframes blinker {
+  50% {
+    opacity: 0;
+  }
+}
+</style>
+
+<div class="components_error">
+  <h1 class="components_error--blinker">CRITICAL FAILURE OMG</h1>
```
illepic: Blerg, I missed this before: let's run all strings through `|t`!

PR #431, opened 2018-10-14
pa11y update

illepic:
- Automated PL urls generation
- pa11y uses pa11y-ci for much simpler config

### DIFF
--- a/tools/pl-paths.js
+++ b/tools/pl-paths.js
@@ -25,4 +25,9 @@ const paths = [].concat(
   )
 );
 
-module.exports = paths.map(path => `patterns/${path}/${path}.html`);
+console.log(paths);
+
+module.exports = {
+  paths,
+  htmlPaths: paths.map(path => `patterns/${path}/${path}.html`),
+};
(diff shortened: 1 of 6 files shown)


tools/pl-paths.js
```diff
@@ -0,0 +1,28 @@
...
+//     // ...
+//   },
+//   atoms: {},
+//   molecules: {},
+//   // ...
+// }
+
+const paths = [].concat(
+  ...Object.values(patternPaths).map(atomic =>
+    Object.values(atomic).map(path => path)
```
atrotter0: I think you can avoid the second map call here: `Object.values(atomic).map(path => path)` if you just access the value directly using `atomic` as the key: `Object.values(patternPaths[atomic])`.

atrotter0: Oh nvm, you are grabbing both `Object.values`.

illepic: I don't think so since the second Object.values() is returning an array of all atomic values, ie `['00-protons-demo-borders', '00-protons-demo-tables', ...]`;

illepic: I want to revisit this.

PR #434, opened 2018-10-17
[WIP, DNM] N design systems

illepic:


mike-potter:
All of the hard-coded "source/default" really bother me, but not sure how else to do it.  Very annoying that YAML doesn't have a way to substitute strings or environment variables.

For Drupal, we could enhance the components module to add a path "prefix" to everything in the *.info.yml file, but that only helps Drupal.

tools/pl-paths.js
```diff
@@ -4,7 +4,7 @@
 
 const {
   patternPaths,
-} = require('../dist/pl/styleguide/data/patternlab-data.json');
+} = require('../dist/app-pl/pl/styleguide/data/patternlab-data.json');
```
mike-potter: Wondering why we need to nest this deeply.  Why both "app-pl" and "pl" and not just "app-pl" itself?

source/default/_patterns/02-molecules/card/demo/cards.twig
```diff
@@ -2,7 +2,7 @@
   {% set container = true %}
 
   {% block column_1 %}
-    <h1>Cards</h1>
+    <h1>Cards flerp</h1>
```
mike-potter: flerp??

jest.config.js
```diff
@@ -12,12 +12,12 @@ module.exports = {
...
-    '^molecules/(.+)': '<rootDir>/source/_patterns/02-molecules/$1',
-    '^organisms/(.+)': '<rootDir>/source/_patterns/03-organisms/$1',
-    '^templates/(.+)': '<rootDir>/source/_patterns/04-templates/$1',
-    '^pages/(.+)': '<rootDir>/source/_patterns/05-pages/$1',
+    '^protons[/]?(.*)': '<rootDir>/source/default/_patterns/00-protons/$1',
+    '^atoms/(.+)': '<rootDir>/source/default/_patterns/01-atoms/$1',
+    '^molecules/(.+)': '<rootDir>/source/default/_patterns/02-molecules/$1',
+    '^organisms/(.+)': '<rootDir>/source/default/_patterns/03-organisms/$1',
+    '^templates/(.+)': '<rootDir>/source/default/_patterns/04-templates/$1',
+    '^pages/(.+)': '<rootDir>/source/default/_patterns/05-pages/$1',
```
mike-potter: Certainly here you want another variable for "source/default"

tools/pl-paths.js
illepic: See structure of app output in dist/: there must be a pl folder to run pl as a sibling to assets.

source/default/_patterns/02-molecules/card/demo/cards.twig
illepic: [WIP]

jest.config.js
illepic: See above comments, give me some air here.

.gitignore
```diff
@@ -1,9 +1,7 @@
 # Compiled files
 dist/
-source/_data/env.json
-source/_patterns/01-atoms/icon/font/
-source/_patterns/01-atoms/icon/scss/_icon-map-generated.scss
-source/_patterns/01-atoms/svgicon/scss/_icons-generated.scss
+env.json
+_icons-generated.scss
```
tjheffner: do these two still need the source/default/ path? I see an env.json in source/default/_data/env.json but not at root

apps/drupal/webpack.config.js
```diff
@@ -49,9 +58,13 @@ const prod = {
...
 module.exports = particle(
+  // app
   { shared, dev, prod },
+  // Default design system
+  APP_DESIGN_SYSTEM,
+  // Use extract css
   {
     cssMode: 'extract',
-    entry: 'app-drupal',
+    entry: APP_NAME, // Called out specifically because 2 entry points
```
tjheffner: is this second entry needed due to the css extraction setup? otherwise it's duplicating the entry in shared, yeah?

particle.js
```diff
@@ -89,12 +90,20 @@ const entryPrepend = entry => ({
...
+ * @param {string} designSystemPath - Webpack config per-design system
  * @param {Object} options - Compile options
  * @param {('hot'|'extract')} options.cssMode - The method of handling CSS output
- * @param {string} options.entry - The main entry point to prepend polyfills
+ * @param {string} [options.entry] - The main entry point to prepend polyfills
  * @returns {*} - Fully merged and customized webpack config
  */
-const particle = (app, options) => {
+const particle = (app, designSystemPath, options) => {
+  // Dynamically pull in design system config. Must be named webpack.config.js
```
tjheffner: this comment about name req should be in all caps or flagged somehow. definitely triple underscores in docs

.prettierignore
```diff
@@ -11,5 +12,5 @@ package-lock.json
 
 _icons-generated.scss
 
-source/_data/env.json
-source/_data/scssVariables.json
+env.json
+scssVariables.json
```
tjheffner: same question about the gitignore here, shouldn't these have full paths still?

tjheffner:
looks like there are eslint & jest-related errors in node 6 also

apps/drupal/webpack.config.js
illepic: image

We have 2 entry points in drupal webpack so that jQuery gets bundled into its own file. The "entry" setting seen in the options sent over to the `particle()` function is which entry point `@babel/polyfill` gets prepended to. In this case we're saying "we need babel polyfill automatically prepended to the big ol' custom blob of javascript, but not the jQuery blob"

particle.js
illepic: Will make sure this is captured in docs, but for now copying a design system folder should work.

PR #437, opened 2018-10-22
Add an example of importing typekit into particle

joannasaerom:


source/_patterns/00-protons/printing/_type.scss
```diff
@@ -7,7 +7,13 @@
 //   webpack.particle.js: sass-loader.data|includePaths
 
 // Example: remote fonts
+// Google Fonts
 @import url('https://fonts.googleapis.com/css?family=Oswald');
+// Adobe Typekit. y0urh45h will need to be updated with your own key.
```
illepic: Go ahead and comment out the import, the comments are perfect.

PR #442, opened 2018-10-25
Add TypeScript as a development option

dslemay:
This brings the ability to use TypeScript within Particle whether in component initialization files, modules, or even in Vue files. It is completely opt in and several files have been move to TypeScript as a reference point.

* Adds TypeScript and associated compiler and linter scripts.
* Webpack now also checks for TypeScript linting and type check errors as well.
* A type-check watch command has also been created for monitoring type safety from the terminal.

Docs on usage within Particle and TypeScript in general are forthcoming and currently being written.


source/default/_patterns/02-molecules/carousel/index.ts
```diff
@@ -21,14 +21,18 @@ export const defaults = {
   interval: 3000,
```
illepic: I can't comment above this line, but shouldn't this be cast/typed to `CarouselSettings`?

source/default/_patterns/02-molecules/vue-widget/src/vue-clock/vue-clock.vue
```diff
@@ -34,31 +34,34 @@
   </div>
 </template>
 
-<script>
-export default {
+<script lang="ts">
```
mkochendorfer: Any reason we converted this entire example over instead of creating a new parallel example with TS?

dslemay: My thinking is that we have several other single file component Vue examples that could be indicative of how to write a Vue SFC in JavaScript. I am currently writing up documentation as well of how to convert a Vue SFC to TypeScript.

source/default/_patterns/02-molecules/vue-widget/src/vue-clock/vue-clock.vue
```diff
@@ -34,31 +34,34 @@
...
     },
-    degrees() {
-      const { seconds, minutes, hours } = this.time;
+    degrees(): TimeInt {
+      const { seconds, minutes, hours }: TimeInt = this.time;
       return {
-        seconds: parseInt((seconds / 60) * 360, 10),
-        minutes: parseInt((minutes / 60 + seconds / 60 / 60) * 360, 10),
-        hours: parseInt((hours / 12 + minutes / 60 / 12) * 360, 10),
+        seconds: (seconds / 60) * 360,
```
illepic: THAAAAAAAAANKS

source/images.d.ts
```diff
@@ -0,0 +1,13 @@
+/* TypeScript requires declaring files which are imported to
```
illepic: Hrm, let's move this and `vue-shims.d.ts` to `tools/` somewhere. With the multi-design system approach, `source/` should have just clean design systems folders right below it. I'm open to other folders/setup for these kinds of files.

webpack.particle.js
```diff
@@ -150,5 +171,6 @@ module.exports = {
       // HTML from the DOM and use it as a template, then remove this line.
       vue$: 'vue/dist/vue.esm.js',
     },
+    extensions: ['.js', '.jsx', '.json', '.ts', '.tsx'],
```
illepic: No .vue here? I need to research `extensions` more.

source/default/_patterns/02-molecules/carousel/index.ts
dslemay: Doesn't need to be. TypeScript recognizes that it is of the same data shape as the interface declared above and infers it. It can be added, but it is an unnecessary annotation as far as the compiler is concerned.

source/default/_patterns/02-molecules/vue-widget/src/vue-clock/vue-clock.vue
dslemay: :+1: Love when type systems remind us what types functions expect to receive and return.

webpack.particle.js
dslemay: Gosh darn it. Missed that one. Will add. Extensions is great because it tells webpack which extensions are resolvable if they are omitted in the import filepath.

source/default/_patterns/02-molecules/carousel/index.ts
dslemay: Going to move it above the export so it is clearer as well.

illepic: So with the interface below this point, TS is cool with this shape from here on?

dslemay: At least within this file. If it were to be needed in another file it would be exported from here and then imported (similar to what is being done in Vue-clock).

illepic:
Looks awesome! I'm holding off on merging to `master` until after we cut to `v10.2`.

source/images.d.ts
dslemay: Moved to tools/typings.

webpack.particle.js
dslemay: Updated.

source/default/_types/particle-element.ts
```diff
@@ -0,0 +1,13 @@
+export default class ParticleElement {
+    constructor(
+        public name: string
+    ) {}
+
+    enable(): any {
```
dslemay: Can the return value from enable and disable be narrowed down further even to a union type? I am not super familiar with the implementation details here.

source/default/_patterns/01-atoms/button/index.js
```diff
@@ -7,6 +7,7 @@ import 'bootstrap/js/dist/button';
 
 // Custom
 import 'protons';
+import ParticleElement from '../../../_types/particle-element.ts';
```
dslemay: If we are going to go down the path of using the class, I think it would be nice to add a modules resolver in the Webpack config to point to `types` as a resolvable module directory. Then at any level of component depth we could do `import ParticleElement from 'types/particle-element'. We also can eliminate the `.ts` since it was added to the resolvable file extensions by webpack.

Here is a link to the webpack modules resolvers as well. https://webpack.js.org/configuration/resolve/#resolve-modules

source/default/_types/particle-element.ts
```diff
@@ -0,0 +1,13 @@
+export default class ParticleElement {
```
dslemay: I think we could also create a ParticleInterface which this class implements to keep all of the type definitions concise and in the interface. It could then also include the types of the arguments that the various functions utilize.

LawlietBlack: Agreed :-) I'll add that

source/default/_patterns/01-atoms/button/index.js
LawlietBlack: Perfect, will do! Thanks for the link too

source/default/_types/particle-element.ts
LawlietBlack: changing it to void, good call

PR #445, opened 2018-10-25
multi design system cleanup

illepic:


source/default/_meta/_00-head.twig
```diff
@@ -13,7 +13,7 @@
 		<!-- End Pattern Lab -->
 
     {% if env == 'production' %}
-    <link rel="stylesheet" href="{{ paths.assets }}/app-pl.styles.css?cacheBuster={{ randomNumber }}">
+    <link rel="stylesheet" href="{{ paths.assets }}/app.styles.css?cacheBuster={{ randomNumber }}">
```
jryanconklin: Not necessarily tied to this PR, but rather than using Faker to generate a random number, would it be better to use Twig's built in Random function? e.g., 
```
<link rel="stylesheet" href="{{ paths.assets }}/app.styles.css?cacheBuster={{ random(10000000) }}">
```

source/default/_meta/_01-foot.twig
```diff
@@ -1,7 +1,7 @@
 	<!--DO NOT REMOVE-->
 	{{ patternLabFoot | raw }}
 
-  <script src="{{ paths.assets }}/app-pl.js?cacheBuster={{ randomNumber }}"></script>
+  <script src="{{ paths.assets }}/app.js?cacheBuster={{ randomNumber }}"></script>
```
jryanconklin: Same as above.

source/default/_meta/_00-head.twig
illepic: @jryanconklin this is a perfect thing I would have never caught. Can you open a quick issue for it?

jryanconklin: On it!

jryanconklin: Oh nice! Looks like you beat me to it.

PR #448, opened 2018-10-25
Adds Enquire.js module

Jwaxo:
Enquire.JS allows for easy usage of breakpoints in order to create both "match" and "unmatch" behavior when resizing or first loading a browser window. I'm a big fan and proponent of it, and it's simple to use. I've added a demo to 00-protons/index.js, and will update the docs accordingly in this PR: https://github.com/phase2/frontend-docs/pull/15


source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,32 @@
 /**
- * Base css generation and global js logic
+ * Base css generation and global js logic.
  */
 
 import './_base.scss';
 
+import { mediaBreakpoint } from 'protons/utilities';
+import scssVariables from '../../_data/scssVariables.json';
```
illepic: The thing that makes this `sassVariables.json` is a Sass plugin, generating JSON meant exclusively for PL to show off. It's straightforward way to get sass variables -> json -> html.

With enquire, can we make it not dependent on this json blob? I'm ok pushing the logic to runtime by reading the `--variables` on `<html>`. The reason I don't want to be dependent on this blob is that Pattern Lab might not be generated in CI when it's time for Drupal to roll around and compile its own assets out of the design system.

source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,32 @@
...
+
+import enquire from '../../../node_modules/enquire.js';
```
illepic: Should just be able to `import enquire from 'enquire.js;`. Let me know if the linter gets pissy about the `.js` on the end of a node_module name.

source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,32 @@
...
+
+export function enable() {
+  // Example usage of the Enquire.js module JS breakpoints.
+  const { GLOBAL_BREAKPOINTS } = scssVariables;
```
illepic: Let's take the approach of reading the CSS variables: https://stackoverflow.com/a/41725782

source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,32 @@
...
  */
 
 import './_base.scss';
 
+import { mediaBreakpoint } from 'protons/utilities';
+import scssVariables from '../../_data/scssVariables.json';
+
+import enquire from '../../../node_modules/enquire.js';
+
+export function enable() {
```
illepic: The "protons" concept doesn't execute an `enable()` function, as it shouldn't need to wait for document.ready(). Just run this logic at root of `protons/index.js`

source/default/_patterns/00-protons/index.js
Jwaxo: Ah, sure. I'm not sure of the best way to ensure that the breakpoints are equal between SCSS and JS, then, other than to just manually configure them (which is, naturally, how I did it in the previous project where I grew to love Enquire).

I'll update to have those defined elsewhere.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic, it's not the linter, I think it's whatever magical namespacing service we use to pull in node_modules. It doesn't get it. I actually used to have a comment here saying as much; my bad for removing it.

illepic: All node modules are available by name, no need for `../../../`.

source/default/_patterns/00-protons/index.js
illepic: Just read those breakpoints off the `--variables` on `<html>`. Those are all set via Sass -> CSS, so the JS would be reading the "single source of truth" here.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic, updated this. Note that it appears running this in proton makes the first "match" not hit until the page gets fully loaded, but at least we have an example where the functionality is defined.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic since the CSS variables were all named in ways that make it impossible to actually loop through them to assign to JS, I've also taken the liberty of making a new CSS variable that holds all of the breakpoint names. This is located in `printing/_elements.css`, but I'm not married to that.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic, this node_module does not load by name. I can't tell you why. Ping me if you want proof.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic, yeah, I just hadn't had the time until I started an hour ago. PR updated.

source/default/_patterns/00-protons/index.js
illepic: image
image

Jwaxo:
Tests are failing because running Enquire in a headless testing environment [historically](https://github.com/akiran/react-slick/issues/93) [causes](https://github.com/WickyNilliams/enquire.js/issues/82) issues.

source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,37 @@
...
+const breakpoints = {};
+
+if (breakpointsString) {
+  breakpointsString.split(' ').forEach(breakpoint => {
+    const value = $('html').css(`--breakpoint-${breakpoint}`);
+    if (value) {
+      breakpoints[breakpoint] = value;
+    }
+  });
+}
```
illepic: There's a really cool chance to use some new tools here. Annotated below.

```js
// Since we'll reference this multiple times, var it
const $root = $(':root');

// We're going to chain down this. It's guaranteed to be at least an empty object.
const breakpoints = $root
  // Read the :root style like before
  .css('--breakpoints')
  // Remove whitespace (see Sass change below)
  .replace(/ /g, '')
  // Split on comma (see Sass change below)
  .split(',')
  // Starting with an empty object (second arg to .reduce()), return a new object
  // that spreads the prior keys into it plus our new key + value
  .reduce(
    (bps, bp) => ({
      ...bps,
      [bp]: $root.css(`--breakpoint-${bp}`),
    }),
    {}
  );
```

source/default/_patterns/00-protons/printing/_elements.scss
```diff
@@ -3,3 +3,14 @@
...
+
+// Combine all breakpoints into a single string.
+$breakpoints: '';
+@each $bp, $value in $grid-breakpoints {
+  $breakpoints: #{$bp} #{$breakpoints};
+}
+
+// Assign breakpoints to --variable at root for reference by Enquire breakpoints.
+:root {
+  --breakpoints: #{$breakpoints};
```
illepic: All you need here is:

```scss
:root {
  --breakpoints: #{map_keys($grid-breakpoints)};
}
```

This gives a comma+space delimited string of breakpoints, no need for `@each`.

Jwaxo: @Illepic, I like this and feel a bit foolish for never using it before. Added in the latest push.

source/default/_patterns/00-protons/index.js
Jwaxo: @illepic This is all great stuff. The one change I had to make was to pull the `.replace` out due to CI running it prior to finding `$root.css('--breakpoints')`, as best I can figure. Bottom line is I was getting "can't split on undefined". So you'll see that difference in the latest push.

source/default/_patterns/00-protons/index.js
```diff
@@ -1,9 +1,37 @@
 /**
- * Base css generation and global js logic
+ * Base css generation and global js logic.
  */
 
 import './_base.scss';
 
+import enquire from 'enquire.js';
+import $ from 'jquery';
+import { mediaBreakpoint } from './utilities';
```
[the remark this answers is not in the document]
Jwaxo: @illepic Aaaaand these are done, too. I like these kind of changes, hahahahaha, personally. STANDARDS.

Jwaxo:
@illepic love all of the changes, I've merged them here just for historical accuracy, and in case anyone wants to see a neater comparison before final merge.

Issue #446, opened 2018-10-25
randomNumber usage should use twig native instead of Faker


illepic:
In files like https://github.com/phase2/particle/blob/f8e5f68a942a54941c2e1a167d857556378c0966/source/default/_meta/_00-head.twig#L16, we really don't need to leverage the data.yml file for stuff like that.

jryanconklin:
Note we're currently using `randomNumber(8)` from Faker to generate a random number. This can be replaced with Twig's built-in random function as random(n). Note that Twig renders random numbers differently than Faker, e.g., `random(n)` will return a number between 0 and the integer param whereas Faker's argument is the decimal place values to return.

Files to Change:
* source/default/_patterns/01-atoms/svgicon/_svgicon.twig
* source/default/_meta/_00-head.twig
* source/default/_meta/_01-foot.twig
* source/default/_data/data.yml

Twig Docs for Random:
https://twig.symfony.com/doc/2.x/functions/random.html

PR #449, opened 2018-10-26
Use random twig function instead of randomNum from faker

atrotter0:
Replaces randomNumber from Faker with twig's random function that generates a random number between 0 and 99999999.

https://github.com/phase2/particle/issues/446


atrotter0:
Do we still want it to create a number that goes up to 8 place values, or can we just call random?

illepic:
@atrotter0 which is faster ;) ?

PR #452, opened 2018-10-26
Update the root url to include app-pl

jengrace:


tools/tests/accessibility/pa11y.js
```diff
@@ -1,7 +1,7 @@
 // All PL paths
 const { htmlPaths } = require('../../pl-paths');
 
-const plRoot = 'http://0.0.0.0:8080/pl'; // @TODO: move this to config
+const plRoot = 'http://localhost:8080/app-pl/pl'; // @TODO: move this to config
```
illepic: Keep the host `0.0.0.0` for now, as this is what is hard-coded in the webpack dev server config for `app/pl/`

PR #454, opened 2018-10-30
Drupal/Grav now read twig from dist/. source/ no longer needs to go to prod.

illepic:
One of the last icky parts of Particle was that `source/` had to be sent along with the **drupal** and **grav** apps because all twig lived in `source/`. After this PR, that is no longer the case.

All `*.twig` referenced within the **drupal**/**grav** app dependency chain is now copied to `dist/app-{drupal,grav}/assets/atomic/`. This means all `demo/` folders and raw assets are "left behind", finally allowing a true dist-only deploy of **drupal** and **grav**.

For example, the only folder structure that would need to go to production for drupal is:

```
apps/
  - drupal/
dist/
  - app-drupal/
```

The only caveat here is that to see changes within drupal/grav while in development `npm run {dev,prod}:{drupal,grav}` needs to be run (because twig is now copied from `source/` to `dist/`.


apps/drupal/webpack.config.js
```diff
@@ -6,25 +6,40 @@ const path = require('path');
...
 const particle = require('../../particle');
 
 // Constants: environment
 const { NODE_ENV } = process.env;
 // Constants: root
-const { PATH_DIST } = require('../../config');
+const { ASSETS_ATOMIC_FOLDER } = require('../../config');
 // Constants: app
-const { APP_NAME, APP_DESIGN_SYSTEM } = require('./config');
+const appConfig = require('./config');
```
dslemay: Should we have these filenames be more descriptive as to what config they are for? It makes sense being in the actual folder, but some of that I think gets lost when removed from the files and they are next to each other as in this case.

dslemay: Even if the file was named appConfig and we could immediately destructure of the resulting require module combining this line and line 19 below.

illepic: I was thinking of moving these file names to something like `particle.app.config.js` to make it clear this is a particle-only file and this is specific to app config. The root file can then become `particle.config.js`.

tjheffner:
it looks like nothing for "npm run start" changes, but what if a client needs pl compiled? should the emitting of twig files to dist be applied to prod pl also? seems odd that only that one would still reference source twig. and if that's the case, does it make more sense to sweep the twig files to one shared folder in dist? e.g. `dist/twig/` etc instead of `dist/app-drupal/` and `dist/app-grav/` and `dist/app-pl`, all containing copies of the same source twig? interested in your thoughts

illepic:
Since the twig files technically still count as part of the Webpack bundle, they're glued to the app-specific output folder (`dist/app-name/assets/`). So there will be duplicates of twig files in output, which is fine since there won't really be Grav + Drupal running at the same time.

I had considered keeping drupal/grav "dev" mode reading from `source/` (no file output), and having only "prod" mode copy files, but that added a ton of extra work and complexity.

apps/drupal/webpack.config.js
dslemay: Looks great.

PR #455, opened 2018-10-30
Add dist/assets path for paths.assets use

Jwaxo:


apps/drupal/particle.theme
```diff
@@ -7,6 +7,15 @@
 
 use Drupal\Core\Form\FormStateInterface;
 
+/**
+ * Implements hook_theme().
+ */
+function particle_preprocess(&$variables) {
+  // Set the paths.assets variable globally, so that drupal can
+  // find webpack-compiled assets.
+  $variables['paths']['assets'] = drupal_get_path('theme', 'particle') . '/../../dist/assets';
```
tjheffner: @Jwaxo does this resolve correctly with the /../../'s appended to the theme path? on previous projects i've hardcoded this to '/themes/custom/particle/dist/assets' but if this works i like it more

illepic: @tjheffner @Jwaxo verifying this works and I'll merge.

PR #456, opened 2018-11-01
Fix Pa11y config and add contrast ratio rules to ignore

jengrace:
Hey @illepic --it appears that in the Pa11y documentation, the list of HTML CodeSniffer rules used is incomplete and missing the .Fail to the end of those contrast rules. Also, after reading further the documentation, I found out that the ignore array needs to be inside of the defaults object.


illepic:
Great work!

PR #463, opened 2018-11-06
Fix remaining accessibility issues

jengrace:
- Added conditional logic to display the alert heading h4 tag only if there is an alert title.
- Anchor name was incomplete. Added the missing "-classes-custom" to the end of it.
- Added conditional logic to display the card title h4 tag only if there is a card title and fixed a typo.
- Originally for the button tag, the type was "button" but to pass the accessibility test, I changed its type to "submit". Also, in order to keep the behavior the same, I added onsubmit="return false" to its form tags.

### DIFF
--- a/source/default/_patterns/02-molecules/vue-widget/demo/vue-html-example.vue
+++ b/source/default/_patterns/02-molecules/vue-widget/demo/vue-html-example.vue
@@ -1,6 +1,7 @@
 <template>
   <form 
-    id="vue-html-example-2" 
+    id="vue-html-example-2"
+    onsubmit="return false"
     class="p-5"
   >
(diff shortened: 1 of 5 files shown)


source/default/_patterns/02-molecules/vue-widget/demo/vue-html-example.vue
```diff
@@ -1,6 +1,7 @@
 <template>
   <form 
-    id="vue-html-example-2" 
+    id="vue-html-example-2"
+    onsubmit="return false"
```
illepic: @jengrace tell me about `return false` and accessibility, I'm learning something new here.

Issue #462, opened 2018-11-06
Updating spacing atom/demo to just use Bootstrap's system


illepic:
The `$spacing-X` vars are a leftover from a prior era. Bootstrap provides what is basically the same thing:

![image](https://user-images.githubusercontent.com/295127/48080002-801a6d80-e1a1-11e8-94c3-a3fdf8453ed9.png)

Let's get our demo in line with this.

illepic:
Another for @jengrace possibly today.

PR #474, opened 2018-11-13
Replace spacing variable with Bootstrap's spacers variable for Spacing demo

jengrace:
Hey @illepic --this is in ref to issue #462 . I replaced the $spacing var with Bootstrap's $spacers Sass map var and then updated $spacing to $spacers where it was needed. Is this what you meant? 

A few more questions:
- Do you want to keep the keys for $spacers as s, m, l or use the default numeric scale in the rest of the system? 
- Do you want the demo to display in rem or px? i.e: 0.3125rem : $spacers(m) **or**  5px : $spacers(m) ? 
- Should we translate the s, m, and l values to rounded and/or whole rem numbers? 
- On the Spacing demo page, it says "To add to these items, adjust $spacing() in _variables.scss" --should we update that to where it says adjust $spacers() in non-printing/_spacing.scss instead?


source/default/_patterns/00-protons/non-printing/_spacing.scss
```diff
@@ -1,7 +1,13 @@
...
-$spacing: (
-  s: 5px,
-  m: 25px,
-  l: 50px,
-);
+$spacer: 1rem !default;
+$spacers: () !default;
+$spacers: map-merge(
+  (
+    s: ($spacer * 0.3125), //5px
```
illepic: Let's not stress putting in our old "s,m,l" versions. Let's just use the existing 1-5.

source/default/_patterns/02-molecules/vue-widget/src/vue-cards/components/banner.vue
```diff
@@ -66,8 +66,8 @@ export default {
 // This is mainly here to show that all variables and mixins are available to
 // Sass here.
 .vue-banner {
-  margin: map-get($spacing, l);
-  padding: map-get($spacing, m);
+  margin: map-get($spacers, l);
```
illepic: Update to use the number vars instead.

source/default/_patterns/00-protons/non-printing/_spacing.scss
```diff
@@ -1,7 +0,0 @@
-// Custom spacing variables and mixins
```
jryanconklin: @illepic 
Consider this PR removes this and replaces with Spacing from Bootstrap, should we be demonstrating an override in Bootstrap Overrides? I don't think so, but want to check in.

source/default/_patterns/00-protons/demo/spacing.twig
```diff
@@ -3,12 +3,15 @@
     <div class="col mt-sm-5">
 
       <h1>Spacing</h1>
-      <p class="lead">To add to these items, adjust <code>$spacing()</code> in <code>_variables.scss</code></p>
-      {% for spacing, value in scssSpacing %}
-        <div style="margin: 10px;">
-          <div style="height: {{ value }}; width: {{ value }}; outline: solid 1px black; display: inline-block; margin-right: 5px;"></div>
-          <span>{{ value }} : <code>$spacing({{ spacing }})</code></span>
-        </div>
+      <p class="lead">To add to these items, adjust <code>$spacers()</code> in Bootstrap's <code>_variables.scss</code>.</p>
```
jryanconklin: If we're using the Bootstrap vars as default, I think we should update from `<code>_variables.scss</code>.</p>` to `<code>_bootstrap_variables.scss</code>.</p>`? CC: @illepic

jryanconklin: Yes! If I'm following how this is changed, I think that's where spacing overrides would now live by using Bootstrap's sizing vars.

tjheffner: Yep! I think this is good to merge once that's updated to `non-printing/_bootstrap-overrides.scss`

jengrace: Awesome. Thanks for clarifying!

source/default/_patterns/00-protons/non-printing/_spacing.scss
jengrace: @jryanconklin @illepic -- I like that idea and added a demo of a few spacing overrides in _bootstrap-overrides.scss. Also, since the overrides must be defined before the imports, I had to use 1rem in place of $spacer. Please review those changes when you get a chance. [496ec3b](https://github.com/phase2/particle/pull/474/commits/496ec3be020c5442ac7e483cd10559d8b15b68e8)

source/default/_patterns/00-protons/non-printing/_bootstrap-overrides.scss
```diff
@@ -16,6 +16,12 @@ $btn-border-radius: 0.25rem;
 // Cards
 $card-border-width: rem-calc(1px);
 
+// Spacing
+$spacers: (
+  6: (1rem * 3.5),
```
illepic: Use font-base-var here.

jengrace: Hey @illepic, I tried replacing `1rem` with `$font-size-base` from `_variables.scss` but got the error `"Undefined variable: "$font-size-base"`. I think it's because the vars are imported after the overrides. I also tried adding the overrides after the @imports but then only the overrides were rendered to the page.

PR #482, opened 2018-11-30
Setting paths via config variables

mike-potter:
NOTE: I also collapsed the build:drupal, build:grav into just build:theme.  We don't have projects that are *both* Drupal and Grav, so they can just override the theme location in .npmrc to point to either Drupal or Grav.  Don't need separate npm commands for both.

That might be a breaking change and haven't looked elsewhere yet to see what that would affect.


jryanconklin:
@mike-potter Have you changed any thoughts on your approach here since this PR was opened?

PR #486, opened 2018-12-03
Fix theme's logo default path

hugonaili:


apps/drupal/particle.theme
```diff
@@ -59,10 +59,11 @@ function particle_theme_suggestions_image_alter(array &$suggestions, array $vari
 function particle_preprocess_block__system_branding_block(&$variables) {
   $logo_default = theme_get_setting('logo.use_default');
   $logo_url = theme_get_setting('logo.url');
+  $logo_svg_inline = drupal_get_path('theme', 'particle') . '/../../dist/app-drupal/assets/logo.svg';
 
   // Set the logo to PL if default.
   if ($logo_default) {
```
jryanconklin: Can we add a better fallback check here in case we bump against this issue again in the future? The logo not existing shouldn't break the site.

PR #490, opened 2018-12-07
Adds Backstop.JS demo configuration and scripts

Jwaxo:


tools/tests/vrt/backstop.json
```diff
@@ -0,0 +1,51 @@
+{
```
tjheffner: Can we use a .js file instead of json here? Something along the lines of: https://github.com/garris/BackstopJS#integration-options-local-install

This will let us get a bit neater with the scenarios array, at least for the _large_ array of potential components to test from PL. I'm envisioning something similar to the pa11y urls, except mapped to a scenario object.

We also should include a mobile viewport with our defaults

Jwaxo: Mmmm I like that, making it work like everything else. I'll look into it.

Jwaxo: @tjheffner Dude good suggestions! I've gone back through and revamped it quite a bit. Let me know if this is... good direction, bad direction, etc!

tools/tests/vrt/backstop-example-settings.js
```diff
@@ -0,0 +1,67 @@
...
+      label: 'article',
+      url: 'http://0.0.0.0:8080/pl/?p=pages-article',
+      ...defaultScenarioSettings,
+    },
+    {
+      label: 'homepage',
+      url: 'http://0.0.0.0:8080/pl/?p=pages-homepage',
+      ...defaultScenarioSettings,
+    },
+  ];
```
tjheffner: i think this is ok to merge now to get the framework of testing in, but we will want to map these urls automatically, eventually.

envisioning something like:

`const { htmlPaths } = require('../../tools/pl-paths.js');
const plRoot = 'http://0.0.0.0:8080/app-pl/pl/';
const scenarios = htmlPaths.map(partial => { label: partial, url: `${plRoot}/${partial}`, ...defaultScenarioSettings });
`

this will get every particle demo page mapped to our default scenario settings, so we don't have to keep adding manually as new components are made. projects will still have to set up additional scenarios manually for dev/staging/prod links, but particle would be covered automatically

tools/pl-paths.js
```diff
@@ -32,8 +33,47 @@ const componentPaths = [].concat(
...
+//           ]
+//         }
+//       ]
+//     }
+//   ]
+// }
+
+// Backstop, or other VRT testing platforms, will use the query forms of the pattern
+// demo paths in order to ensure all styles are present and rendered.
+const partials = [].concat(
```
illepic: Opportunity to use `.reduce()` and destructuring:

```
const partials = navItems.patternTypes.reduce((acc, { patternTypeItems }) => {
  patternTypeItems.forEach(({ patternSubtypeItems }) => {
    patternSubtypeItems.forEach(({ patternName, patternPartial }) => {
      patternName !== "View All" && acc.push(patternPartial);
    });
  });
  return acc;
}, []);
```

tools/tests/vrt/backstop-example-settings.js
Jwaxo: For the record, these comments have been implemented.

tools/tests/vrt/backstop-example-settings.js
```diff
@@ -0,0 +1,66 @@
+/**
```
illepic: NICE

Issue #425, opened 2018-10-11
Integrate Drupal Layout API in Drupal App


jryanconklin:
### Task Description

Integrate default layouts for registry via Layout Discovery module in Particle's Drupal app. See Drupal's docs for information on the Layout API:

https://www.drupal.org/docs/8/api/layout-api

**Note** this work is separate from using the Layout Builder module, but is a step toward using this by default in Particle.

### Acceptance Criteria

* Layouts are registered to Drupal on theme installation
* There's several layouts to choose from including 1 column, 2 column and 3 column that account for Particle's current regions (see regions key in https://github.com/phase2/particle/blob/master/apps/drupal/particle.info.yml)
* A Particle user can switch between regions on default content types (article and basic page).

#### References
https://www.unleashed-technologies.com/blog/2017/11/05/are-you-leveraging-drupal-8s-layout-api
https://www.ostraining.com/blog/drupal/new-layout-builder/
https://www.phase2technology.com/blog/creating-mega-menu-layout-builder

PR #491, opened 2018-12-07
Add Drupal layouts to Drupal app

jengrace:
#425
 
This PR adds 1 - 4 column default layouts to the Drupal app and are registered via the `particle.layouts.yml` file.

### DIFF
--- a/apps/drupal/particle.layouts.yml
+++ b/apps/drupal/particle.layouts.yml
@@ -1,7 +1,7 @@
 particle_one_col:
-  label: 'Particle One Column'
-  path: layouts/test-onecol
-  template: test-onecol
+  label: 'Particle one column'
+  path: layout/particle-onecol
+  template: particle-onecol
   category: 'Columns: 1'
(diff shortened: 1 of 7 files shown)


jengrace:
Thanks for the reviews! A few additional changes were made. After conversation with @illepic, we decided to set the container to false for each layout and leave it up to the user if they want to add an additional container around the layout.

apps/drupal/particle.layouts.yml
```diff
@@ -0,0 +1,56 @@
+particle_one_col:
+  label: 'Particle one column'
+  path: layout/particle-onecol
+  template: particle-onecol
+  category: 'Columns: 1'
```
illepic: Is `category` something we could leverage? ie a "Particle" category?

jengrace: @illepic --yes, I believe so! To confirm, do you mean replace "Columns: 1" with "Particle"?

PR #516, opened 2019-02-28
update paths.assets for drupal, update default pa11y config

tjheffner:
should fix the issue where paths.assets doesn't work out of the box for drupal anymore (due to how the dist folder structure has been updated)

### DIFF
--- a/tools/pl-paths.js
+++ b/tools/pl-paths.js
@@ -20,11 +20,20 @@ const {
 //   // ...
 // }
 
+// all particle paths
 const paths = [].concat(
+  ...Object.values(patternPaths).map(atomic => Object.values(atomic))
+);
+
+// only check our actual components from atoms up
+delete patternPaths.protons;
+delete patternPaths.atoms.svgicons;
+const componentPaths = [].concat(
   ...Object.values(patternPaths).map(atomic => Object.values(atomic))
 );
 
 module.exports = {
   paths,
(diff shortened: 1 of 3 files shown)


tools/pl-paths.js
```diff
@@ -20,11 +20,20 @@ const {
...
+// all particle paths
 const paths = [].concat(
   ...Object.values(patternPaths).map(atomic => Object.values(atomic))
 );
 
+// only check our actual components from atoms up
+delete patternPaths.protons;
+delete patternPaths.atoms.svgicons;
+const componentPaths = [].concat(
+  ...Object.values(patternPaths).map(atomic => Object.values(atomic))
```
Jwaxo: Check this thing out.

tjheffner: hah, it's essentially a carbon copy of the "paths" variable above, just with some other things i don't necessarily care about for accessibility testing removed. original paths is still available for VRT testing

tools/pl-paths.js
```diff
@@ -20,11 +20,20 @@ const {
 //   // ...
 // }
 
+// all particle paths
 const paths = [].concat(
   ...Object.values(patternPaths).map(atomic => Object.values(atomic))
 );
 
+// only check our actual components from atoms up
```
illepic: Lets' get a config array for "delete these things before pa11y testing".

PR #521, opened 2019-03-12
Fix issue of breaking layout when changing css in Chrome Devtools

hugonaili:


illepic:
Just want to verify that you can still see the Sass file referenced from CSS in dev mode with this change.

PR #523, opened 2019-03-18
Update Drupal App with Theme Tools and Theme File Organization

jryanconklin:
This PR adds:

* ParticleTools: This is a simple Class based approach to adding theme tools (helper functions and custom services) to Particle. An optional improvement for code organization / optimization.
* Theme Hook Includes: Theme hooks have been organized into the `includes` directory to match work done on other internal projects. Offers code org scheme for hooks to prevent `.theme` file explosion.

### DIFF
--- a/apps/drupal/includes/blocks.theme.inc
+++ b/apps/drupal/includes/blocks.theme.inc
@@ -1,0 +1,44 @@
+<?php
+
+/**
+ * @file
+ * Functions to support theming blocks in the Particle theme.
+ */
+
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function particle_theme_suggestions_block_alter(array &$suggestions, array $variables) {
+  // Provide suggestion for block templates by custom block type.
+  if (!empty($variables['elements']['content']['#block_content'])) {
+    $block = $variables['elements']['content']['#block_content'];
+    // Add `block--BLOCK-TYPE.html.twig`.
+    $suggestions[] = 'block__' . $block->bundle();
+    $view_mode = $variables['elements']['#configuration']['view_mode'];
+    if (!empty($view_mode)) {
+      // Add `block--BLOCK-TYPE--VIEW-MODE.html.twig`.
+      $suggestions[] = 'block__' . $block->bundle() . '__' . $view_mode;
+    }
+  }
+}
+
+/**
+ * Implements hook_preprocess_block().
+ */
+function particle_preprocess_block__system_branding_block(&$variables) {
+  $logo_default = theme_get_setting('logo.use_default');
+  $logo_url = theme_get_setting('logo.url');
+
+  $logo_svg_theme = drupal_get_path('theme', 'particle') . '/logo.svg';
+  $logo_svg_assets = drupal_get_path('theme', 'particle') . '/../../dist/app-drupal/assets/logo.svg';
+  $logo_svg_inline = file_exists($logo_svg_assets) ? $logo_svg_assets : $logo_svg_theme;
+
+  // Set the logo to PL if default.
+  if ($logo_default) {
+    $variables['logo_svg_inline'] = $logo_svg_inline;
+    // Send to img if Logo is not an default.
+  }
+  else {
+    $variables['logo_img_src'] = $logo_url;
+  }
+}
--- a/apps/drupal/includes/blocks.theme.inc
+++ b/apps/drupal/includes/blocks.theme.inc
@@ -5,6 +5,8 @@
  * Functions to support theming blocks in the Particle theme.
  */
 
+use Drupal\particle\Particle;
+
 /**
  * Implements hook_theme_suggestions_HOOK_alter().
  */
@@ -29,8 +31,8 @@ function particle_preprocess_block__system_branding_block(&$variables) {
   $logo_default = theme_get_setting('logo.use_default');
   $logo_url = theme_get_setting('logo.url');
 
-  $logo_svg_theme = drupal_get_path('theme', 'particle') . '/logo.svg';
-  $logo_svg_assets = drupal_get_path('theme', 'particle') . '/../../dist/app-drupal/assets/logo.svg';
+  $logo_svg_theme = drupal_get_path('theme', Particle::THEME_NAME) . '/logo.svg';
+  $logo_svg_assets = drupal_get_path('theme', Particle::THEME_NAME) . '/../../dist/app-drupal/assets/logo.svg';
   $logo_svg_inline = file_exists($logo_svg_assets) ? $logo_svg_assets : $logo_svg_theme;
 
   // Set the logo to PL if default.
--- a/apps/drupal/includes/blocks.theme.inc
+++ b/apps/drupal/includes/blocks.theme.inc
@@ -32,7 +32,7 @@ function particle_preprocess_block__system_branding_block(&$variables) {
   $logo_url = theme_get_setting('logo.url');
 
   $logo_svg_theme = drupal_get_path('theme', Particle::THEME_NAME) . '/logo.svg';
-  $logo_svg_assets = drupal_get_path('theme', Particle::THEME_NAME) . '/../../dist/app-drupal/assets/logo.svg';
+  $logo_svg_assets = Particle::ASSETS_PATH . '/logo.svg';
   $logo_svg_inline = file_exists($logo_svg_assets) ? $logo_svg_assets : $logo_svg_theme;
 
   // Set the logo to PL if default.
--- a/apps/drupal/includes/form.theme.inc
+++ b/apps/drupal/includes/form.theme.inc
@@ -1,0 +1,70 @@
+<?php
+
+/**
+ * @file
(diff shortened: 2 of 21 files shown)


apps/drupal/src/Tools/ParticleTools.php
```diff
@@ -0,0 +1,38 @@
+<?php
+
+namespace Drupal\particle\Tools;
+
+use Drupal\particle\Particle;
+
+/**
+ * Provides utility functions for all preprocessors.
+ *
+ * This approach is influenced by Adventist's theme tools system.
```
illepic: Nuke name.

apps/drupal/src/Particle.php
```diff
@@ -0,0 +1,29 @@
...
+   * Provides theme name constant.
+   *
+   */
+  const THEME_NAME = 'particle';
+
+  /**
+   * Provides the path for compiled Drupal assets in build system.
+   *
+   */
+  const ASSETS_PATH = '/../../dist/app-drupal/assets';
```
illepic: Is there any recommended way to relative pathing that isn't just `../../../`

apps/drupal/src/Tools/ParticleTools.php
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  /**
+   * Return the theme's assets path relative to the Drupal root.
+   *
+   * @return string
+   *   The theme's assets path.
+   */
+  public static function getAssetsPath() {
+    return static::getThemePath() . Particle::ASSETS_PATH;
```
illepic: Let's use pathing util functions for this.

apps/drupal/src/Tools/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+## Usage
+
+If you have a helper function you'd like to include in your app, 
+simply add the functionality to `ParticleTools.php`. 
+
+Particle Tools have already been added to individual `theme.inc` files, however if you'd 
+like to use the Tools elsewhere in PHP simply include the namespace to the file:
+
+`use Drupal\particle\Tools\ParticleTools;`
```
illepic: Triple tics fo' lyfe

apps/drupal/includes/misc.theme.inc
```diff
@@ -0,0 +1,24 @@
...
+  // Set the paths.assets variable globally, so that drupal can
+  // find webpack-compiled assets.
+  $variables['paths']['assets'] = ParticleTools::getAssetsPath();
+}
+
+/**
+ * Implements hook_preprocess_user().
+ */
+function particle_preprocess_user(&$variables) {
+  $variables['content']['member_for']['#markup'] = '';
```
illepic: What dis.

apps/drupal/includes/layout.theme.inc
```diff
@@ -0,0 +1,16 @@
...
+
+/**
+ * @file
+ * Functions to support theming page layout in the Particle theme.
+ */
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function particle_preprocess_page(&$variables) {
```
illepic: Suggestion: Move to Particle tools.

apps/drupal/includes/blocks.theme.inc
```diff
@@ -0,0 +1,46 @@
...
+
+/**
+ * Implements hook_preprocess_block().
+ */
+function particle_preprocess_block__system_branding_block(&$variables) {
+  $logo_default = theme_get_setting('logo.use_default');
+  $logo_url = theme_get_setting('logo.url');
+
+  $logo_svg_theme = drupal_get_path('theme', Particle::THEME_NAME) . '/logo.svg';
+  $logo_svg_assets = drupal_get_path('theme', Particle::THEME_NAME) . '/../../dist/app-drupal/assets/logo.svg';
```
illepic: Use our utils here.

apps/drupal/includes/blocks.theme.inc
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * Implements hook_preprocess_block().
+ */
+function particle_preprocess_block__system_branding_block(&$variables) {
+  $logo_default = theme_get_setting('logo.use_default');
+  $logo_url = theme_get_setting('logo.url');
+
+  $logo_svg_theme = drupal_get_path('theme', 'particle') . '/logo.svg';
+  $logo_svg_assets = drupal_get_path('theme', 'particle') . '/../../dist/app-drupal/assets/logo.svg';
```
mike-potter: Should not hard-code the path to "dist" here.  Or at least add a TODO: since Particle should have a config for where "dist" is located.

apps/drupal/includes/form.theme.inc
```diff
@@ -0,0 +1,70 @@
...
+ * Implements hook_theme_suggestions_HOOK_alter().
+ */
+function particle_theme_suggestions_form_alter(array &$suggestions, array $variables) {
+  $id = $variables['element']['#id'];
+  // Particle Theme for Octane's Views Search and Drupal's Built-In Search.
+  // This hook allows to apply two search block implementations to the same
+  // theme file. Drupal only template file located at
+  // /particle/apps/drupal/templates/form/form--particle-search.html.twig.
+  if ($id == 'views-exposed-form-search-page' || $id == 'search-block-form') {
+    $suggestions[] = 'form__particle_search';
```
mike-potter: Don't think "particle" name should be used here.  Projects have their own theme name (renamed from Particle) so would prefer a non-specific name more related to function here.

apps/drupal/includes/navigation.theme.inc
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * @file
+ * Functions to support theming navigation in the Particle theme.
+ */
+
+/**
+ * Implements hook_preprocess_breadcrumb().
+ */
+function particle_preprocess_breadcrumb(&$variables) {
+  if (($node = \Drupal::routeMatch()->getParameter('node')) && $variables['breadcrumb']) {
```
mike-potter: Need comment here on what this preprocess is doing.

apps/drupal/src/ParticleTools/ParticleTools.php
```diff
@@ -0,0 +1,36 @@
+<?php
+
+namespace Drupal\particle\ParticleTools;
+
+/**
+ * Provides utility functions for all preprocessors.
+ *
+ * This approach is influenced by Adventist's theme tools system.
```
mike-potter: Do we really want to reference a specific client name here?

apps/drupal/src/ParticleTools/ParticleTools.php
```diff
@@ -0,0 +1,36 @@
...
+ * For advanced examples including implementing custom services,
+ *
+ * @see https://bitbucket.org/phase2tech/adventist/src/develop/src/themes/ahs_theme/src/ThemeTools.php
```
mike-potter: Don't put a link to a private bitbucket in the public particle project.

mike-potter: Would rather see the approach here documented somewhere in the Particle docs and linking to that.  Also a basic summary of what this ParticleTools is for.

mike-potter: Oh, and I see there is a README file below, so just link to that and put docs about "advanced examples" there.

mike-potter:
Sorry, looking at my code review comments I realized later they are more about the existing Particle theme code.  I see that this PR just split that existing code into modules, which is good.  So those comments should probably go to the wider Particle team for review.

On the actual approach here, I don't want to see this added to Particle theme directly as "apps/drupal/src" files.  ParticleTools needs to be put into an actual Drupal module and then brought into a project via Composer.  If this is just an "example of what a project could do" then it should still be a separate example module.  

Keep in mind that projects using Particle will almost always have their own theme name (and some projects will have multiple themes).  A script/generator to rename the Particle theme to something else will be needed.  So anything that contains the theme name makes this task more complex.  Try to modularize and name based on the functionality being added more generically and not the specific "Particle" theme name.  Think about how something like ParticleTools would work in a multi-theme site as a more generic module that is driven by config settings and not anything hard coded.  Projects will potentially have different directory structures defined in config files.

jryanconklin:
@mike-potter Looking through your comments on what was merged from this PR, I think most of your concerns were addressed. I think it's appropriate to use the particle name in initial theme hooks since we want a demo-able version out of the box. I really don't expect our internal projects to keep these theme hooks around. 

I did think of the challenge of multi-design projects and theme renaming here and decided it was more worth "try this out and decide what's best" rather than attempting to nail it in one. I think some iterations will be needed to get what we want here. To me theme functions belong inside the theme and Tools are your theme functions when you've split your `.theme` file into separate includes. But I direct to @illepic here on questions of approach since it's his baby.

apps/drupal/includes/form.theme.inc
jryanconklin: This is what's currently in Particle. I think we need to consider all default theme hooks as "this makes a demo-able version on install and is intended to be deleted."

Issue #503, opened 2019-01-02
Namespaces from an app do not necessarily correlate to directories generated for that app


illepic:
See: https://www.drupal.org/project/components/issues/2993671#comment-12912396

Our design system provides a lot of namespaces. Drupal then comes around and references most of those namespaces ... but not necessarily all.

We need to ensure that the namespaces being written out to places like `particle.info.yml` lines up with what is actually used within those apps.

In the instance referenced in the Drupal issue above, the card-grid organisms was never referenced within the Drupal build, therefore it wasn't copied to the `dist/app-drupal/atomic/` directory. This caused the Component Libraries Drupal module to throw out the entire `@organisms` namespace.

jryanconklin:
This issue was duplicated in issue 512. Closing that issue and copying text here:

## Problem / Motivation
The alpha version of the Component Library module no longer supports `component-libraries` that have unresolvable keys. A provided key that does not have a directory in `apps/dist` will cause a whitescreen. 

## Proposed Solution
We should provide a check in namespace maker to guarantee we only update keys with directories that are in `dist`.

Nebel54:
I added a fix for this issue here: #525

jryanconklin:
Resolved in #552

PR #525, opened 2019-03-20
Prevent creation of non-existing namespaces in apps

Nebel54:
As stated in #503 Drupal throws an error when using the current Component Library module. To prevent this, I needed to:

- Move the generation of the namespaces after the compile process (from an npm run script to a webpack `RunScriptAfterEmit` plugin), so the generator can check if all generated namespaces exist in dist/
- Change the existing namespace transform functions a bit to also return the relative path for the file checks.
- Add another flag `skipFileExistenceCheck`, to allow different design systems to skip this file check, as patternlab needs all components of course.

I am not quite sure if the namespace generation really needs to run for patternlab before the webpack build, but I just left it in `gulp compile:startup`, so it runs before and after the webpack build now.


illepic:
Will take a look at this!

illepic:
Noting here that I haven't forgotten about this. Circling around to it this week.

jcandan:
I can vouch for this patch; this fixed the `Uncaught PHP Exception Twig\Error\LoaderError` I was getting. See #543.

lunitin:
This PR also fixed my issue with Drupal 8.7.1

> I can vouch for this patch; this fixed the `Uncaught PHP Exception Twig\Error\LoaderError` I was getting. See #543.

This PR also fixed the same issue that cropped up with my Drupal 8.7.1 install.

Nebel54:
As a quick update: This PR will not make it into particle. PR #552 already provides a short-term solution and in #554 will contain a cleaner long-term-solution.

PR #527, opened 2019-03-27
Add Preprocessor for Particle Drupal Theme

jryanconklin:
This PR:
* Updates ParticleTools for getTheme method
* Adds experimental Preprocessor tool


apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,432 @@
...
+  }
+
+  /**
+   * Get the preprocessor class map.
+   *
+   * @return array
+   *   An array with keys of preprocess name and values of full class path.
+   */
+  public static function getPreprocessorsClassMap() {
+    if (!isset(static::$preprocessorClassMap)) {
```
recrit: @jryanconklin This cannot be a one-time cached variable. This uses static::$allSuggestions which may not be built yet or might be incomplete. Example: First call to this is for node.card and static::$allSuggestions is populated with suggestions for node in a card view mode. The next call is for a node in hero view mode (or even a different entity type - paragraph), then the static::$preprocessorClassMap would not get updated since it already was set which forces node.hero to fallback to the "default" class.

recrit:
@jryanconklin Thanks! This is a good clean start for a new project.  
I added a comment on the diff for how the statically cached variables are being treated.

jryanconklin:
I've updated this to use the State API + `drupal_static` to handle the preprocessor class map.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,500 @@
...
+  }
+
+  /**
+   * Get the preprocessors from directory and subdirectories.
+   *
+   * @return array
+   *   An array of preprocessors from the file system.
+   */
+  public static function getPreprocessors() {
+    $preprocessors = [];
```
recrit: @jryanconklin Code suggestion: Cache $preprocessors with a static class property or drupal_static() so that all directories are not scanned each time new suggestions are found for mergedMaps().

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,500 @@
...
+      $map['default'] = __NAMESPACE__ . '\\Preprocessors\\Base';
+      drupal_static_reset();
+      \Drupal::state()->set(static::STATE_ID, $map);
+      $preprocessor_map = static::getPreprocessorsClassMap();
+    }
+
+    // If we have new suggestions, try to update the preprocessor map state.
+    if (!empty($new_suggestions)) {
+      $new_maps = static::mergedMaps($new_suggestions, $preprocessor_map);
+      if (!empty($new_maps) && $new_maps !== $preprocessor_map) {
```
recrit: @jryanconklin Both arrays need to be sorted by **value** before checking for equality in PHP. This could be done in the getPreprocessorsClassMap() and mergedMaps().
Example:
`
$x = ['a' => 1, 'b' => 2];
$y = ['a' => 1, 'b' => 2];
$z = ['b' => 2, 'a' => 1];

$x === $y; // returns TRUE
$x === $z; // returns FALSE

`

jryanconklin: Updated! Added an asort which should ensure these are in the same state.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: Updated! Thanks for the suggestion!

jryanconklin:
I've updated the file directory scanner to use Drupal's built in method. 

It still has some manipulation to do to guarantee that Base isn't included and that we give the entity or page type as the value of the array (needed to map by vars). We could evaluate moving Base outside this folder and giving it it's own namespace since it's manually created as the default.

apps/drupal/src/Plugin/Preprocessor/README.md
```diff
@@ -0,0 +1,51 @@
...
+ * Implements template_preprocess_page().
+ */
+function particle_preprocess_page(&$variables) {
+  // Process All Variables Via PreprocessClass.
+  PreprocessorManager::loadByVariables('page', $variables)
+    ->process($variables);
+}
+```
+
+With the above in place as the hook, `loadByVariables` will direct hook type
```
mikecrittenden: That makes sense for this simple example, but what about nodes of specific type and display modes? Do I need to write the preprocess myself to pull those files in, or are you saying it happens automatically just by naming convention? If that's the case, why is it needed manually for Pages?

apps/drupal/src/Plugin/Preprocessor/README.md
```diff
@@ -0,0 +1,51 @@
...
+* `Base.php`
+* `Page.php`
+* `Node.php`
+
+The `Base.php` preprocessor is the root implementation all preprocessors inherit from.
+It represents the base implementation of a preprocessor.
+
+`Page.php` and `Node.php` preprocess their respective hooks via the PreprocessManager method:
+
+See `particle_preoprocess_page` hook to see how this is added:
```
mikecrittenden: Typo: preoprocess

Also, I don't see `particle_preprocess_page()` anywhere in this PR so this line is confusing to me. Does the snippet below actually exist somewhere or is it just in this README as an example?

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,509 @@
+<?php
```
mikecrittenden: This seems like a lot of code for it to need to exist per-Drupal-theme. Should it be moved to a contrib module which Particle depends on (I can see why you'd hate having to do that, but just for the sake of discussion I wanted to ask it) or at least moved out of the individual app so that if there are multiple Drupal themes they can all reference the same file instead of each having copies?

Or is this file meant to be owned/altered by the project, in which case it makes sense for it to be here? But that would open up a whole new discussion - that this is a lot of pretty deep code to ask each project to understand and maintain.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
```diff
@@ -0,0 +1,20 @@
+<?php
+
+namespace Drupal\particle\Plugin\Preprocessor\Preprocessors\Node;
+
+use Drupal\particle\Plugin\Preprocessor\Preprocessors\Base;
+
+/**
+ * Provides a base node preprocessor.
+ */
+class Node extends Base {
```
mikecrittenden: Why doesn't this one have a `process()` function?

apps/drupal/src/Plugin/Preprocessor/README.md
```diff
@@ -0,0 +1,51 @@
...
+function particle_preprocess_page(&$variables) {
+  // Process All Variables Via PreprocessClass.
+  PreprocessorManager::loadByVariables('page', $variables)
+    ->process($variables);
+}
+```
+
+With the above in place as the hook, `loadByVariables` will direct hook type
+preprocessing to the correct folder and type. In `Preprocessors/Page.php` you
+can now update your variables in the `process` method.
```
mikecrittenden: Maybe something that would help with general "how do I use this?" would be a very specific step by step guide in this README with an example. Like:

----

Here's how to create a new variable for the Twig template for a node type called Article, for the display mode Teaser:

1. Create NodeArticleTeaser.php inside blah
2. Copy/paste the contents of Node.php into it and rename the class
3. Blah blah blah

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,509 @@
...
+ *
+ */
+class PreprocessorManager {
+
+  /**
+   * The State API ID for this Manager.
+   *
+   * @var string
+   */
+  const STATE_ID = 'preprocessor.map';
```
mikecrittenden: This seems a little tricky if trying to support multiple Drupal themes, or use base themes. The state() for each of them would clobber each other unless the developer knows to manually prefix this with the theme name or whatever. I can get that maybe "support multiple themes" would be a follow up issue, but it seems like at least we should try to not cause bugs for multiple themes, even if they aren't necessarily supported.

Probably not a problem for most sites since they wouldn't be using multiple themes on the same site - it would be a multisite scenario in which case Drupal would probably only see the one that exists in the theme that's active on that site. But still just something to note/be aware of.

Maybe it should just be prefixed by `ParticleTools::getThemeName()`?

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,509 @@
...
+      // Entity preprocessing.
+      $entity = NULL;
+      $entity_type = NULL;
+      $entity_bundle = NULL;
+      $entity_group_bundle = NULL;
+      $entity_view_mode = NULL;
+
+      if ($hook === 'page') {
+        // Page: Find the page entity.
+        $route_entity = static::getPageEntity();
```
mikecrittenden: Probably out of scope for this PR but have you looked at if/how this could be used for preprocessing Views? Seems like this mostly relies on the page rendering a content entity and Views are config entities. Not a big deal but it would be cool if `template_preprocess_views_view()` could be a thing of the past.

apps/drupal/src/Plugin/Preprocessor/README.md
```diff
@@ -0,0 +1,51 @@
...
+
+The Preprocessor directory contains the PHP work to preprocess templates.
+To match the Plugin system in Drupal core, these files should be named in the following manner:
+
+Entity - Bundle - View Mode 
+
+The Preprocessor directory is automatically scanned and loaded as a Class Map in ParticlePreprocessorManager.php.
+
+For now, each entity, bundle and view name should be separated by a capital letter.
+For example, if you have a `node--article--card-max`, name your php file: `NodeArticleCardmax.php`
```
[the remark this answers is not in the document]
jryanconklin: Anything you'd put into a theme preprocess function, e.g., `particle_preprocess_node__article()`, with the benefit that you can:
* Progressively build up those alterations all the way up the View mode level because inheritance
* Use xDebug inside these files.
* Organize them by their types (rather than have a ton of hooks)
* Easily export the pattern you've preprocessed in named chunks rather than relying on `content|without()` in twig

apps/drupal/src/Plugin/Preprocessor/README.md
jryanconklin: Each type is only "turned on" by placing it on the hook you want to use it for. Because it's essentially mapping the hooks and their theme suggestions to the classes you create and rerouting them to that map. `loadByVariables` here says, "okay give me all your page suggestions, and load them into a class if it's available (e.g., a file with that name exists)."

apps/drupal/src/Plugin/Preprocessor/README.md
jryanconklin: Ah, yeah, I removed that and will update the README for the PR. Basically didn't want anywhere where we turned this on by default, since it should 100% be optional.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
jryanconklin: It's not needed here since it'd just be empty. But if you wanted to only preprocess a Node, you'd add that via  `process` here.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: To me it makes most sense for the theme and it's tools to own their implementation. But I don't disagree in that this file is a bit much. I think this ought to be broken up and tuned. But definitely wanted v1 to be a fairly direct port, and then open up to us at large to tune it.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: This is a great suggestion!

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: For sure. Mike did some much more complex uses of this and I'd love to see it expand and get tuned.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
mikecrittenden: But Page has a preprocess function that isn't doing anything? Are you assuming that Page probably will need some processing and Node problem won't?

apps/drupal/src/Plugin/Preprocessor/README.md
mikecrittenden: Maybe it would make sense for particle.theme to have the preprocess_page and preprocess_node implementations in place, but commented out, so that it's a bit more obvious how to start using it?

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: @mikecrittenden A class constant cannot contain an expression. So workaround here would be to:
* Prefix the theme name in the implementation of the State API (easiest)
* Expect devs to change this in multi-theme

mikecrittenden: This is not at all a PR blocker anyway - feel free to ignore and we can address when an actual example of it causing a problem appears.

apps/drupal/src/Plugin/Preprocessor/README.md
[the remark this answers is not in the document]
jryanconklin: @mikecrittenden Okay. I updated the README. Let me know if that makes it a bit more clear.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
jryanconklin: Fair. I added it to Node. It's probably better that way as an example.

apps/drupal/src/Plugin/Preprocessor/README.md
jryanconklin: I added how to turn it on for Node and Page to the README. Figured that'd be easiest!

apps/drupal/src/Plugin/Preprocessor/README.md
jryanconklin: Updated README.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: Added a note to the README. I think there's easy ways to handle this, but I think that should be a discussion within the project that _wants_ to use Preprocessor and multi-design. But really good to think about early and often!

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
mikecrittenden: I don't see it?

jryanconklin: Ack! Put it in the wrong spot. Updated!

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,509 @@
...
+   * Get the preprocessors from directory and subdirectories.
+   *
+   * @return array
+   *   An array of preprocessors from the file system.
+   */
+  public static function getPreprocessors() {
+    $preprocessors = &drupal_static(__FUNCTION__);
+    if (is_null($preprocessors)) {
+      // Get all files in the preprocessors directory.
+      $files = file_scan_directory(static::getPreprocessorsDirectory(), '/.php/');
```
jhedstrom: I wonder if these can be further cached? I'd be worried about potential performance of doing a file scan on each page load...

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
```diff
@@ -0,0 +1,28 @@
...
+   * {@inheritdoc}
+   */
+  public function init(array &$variables) {
+    parent::init($variables);
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function process(array &$variables) {
```
jhedstrom: While I think it's valuable to have this here as an example, I wonder if the above `init` function needs to be duplicated here, since that is unlikely to be customized per processor?

jhedstrom:
Note, here's docs on how to allow [plugin discovery inside of themes](https://www.drupal.org/docs/8/api/plugin-api/creating-a-plugin-that-can-be-defined-in-themes).

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
jryanconklin: Since this is only ever called in mergedMaps, which is behind a check for new suggestions, I think it's okay. This should only run when there's something to update, which seems appropriate to me. But absolutely agree a better caching scheme might be good! But I think we'll need some evidence from actually using this on a project to move forward.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
```diff
@@ -0,0 +1,509 @@
...
+use Drupal\Core\Entity\EntityInterface;
+use Drupal\particle\Tools\ParticleTools;
+
+/**
+ * Provides a theme preprocessor manager for Class based theme preprocessors.
+ *
+ * Drupal core does not provide built-in plugin support for theme files.
+ * The plugin discovery might before the front end theme is not active.
+ *
+ * The approach here is to use Drupal's State API to handle Preprocessor Class.
```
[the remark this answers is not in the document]
jryanconklin: Possibly. I chose the State API because we need the map as a key value pair store and it was something I was comfortable using.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Node/Node.php
jryanconklin: Removed.

apps/drupal/src/Plugin/Preprocessor/Preprocessors/Base.php
```diff
@@ -0,0 +1,115 @@
...
+  public function process(array &$variables) {
+    $this->ensure($variables);
+    return $this;
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  public function ensure(array &$variables) {
+    if (empty($variables['_init_processed'])) {
```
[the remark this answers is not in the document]
jryanconklin: This is leftover from the original version. I think it's good like this as just a demo thing, which is why I left it in. But if it's distracting I'm happy to remove it.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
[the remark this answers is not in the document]
mike-potter: I agree that this should be using the Drupal cache API and not the State API.  State is not intended for this kind of caching and does not benefit from other cache services or follow with the normal cache clearing systems.  The Cache system is very flexible and shouldn't be difficult to learn or refactor.

apps/drupal/src/Plugin/Preprocessor/PreprocessorManager.php
mike-potter: I think this is another area where the normal Cache API can be used.  A file scan in a targeted directory like this might not seem bad, but I've seen several client instances with slow file systems where any sort of file access needed to be avoided and cached.  I feel like the actual performance is less of an issue than the architecture of this.  And again it's a perfect use-case for a cache bin.

mike-potter:
I am super excited about the concept here, but it definitely needs some work before I'd be comfortable seeing it as part of Particle or used across P2 projects.

I agree with Jonathan that this really belongs in a module.  Some things not already mentioned are:

- A module would allow participation in Drupal security policy
- A module would allow version (updates, dependencies, etc) management via composer
- With a module, a site with multiple themes would have a single copy of the preprocessor scaffolding rather than copies within each theme folder. 
- Putting this code into an obscure theme on github doesn't fit with the community model of Drupal and how P2 supports this community.

jryanconklin:
Closing PR for now in light of the requested changes. Also plan to evaluate Preprocess module. Thanks for everyone that checked it out and commented!

PR #535, opened 2019-04-16
Update bootstrap spacer override and better document usage

jryanconklin:
This PR refactors our Bootstrap approach for spacers. Rather than simply append, override the entirety of the list to better show off what's going on. This change also includes a blurb in the demo proton to give devs a better idea how to use this in place of static pixel values.

Note that this may negatively impact some default patterns. Any that previously used spacers 5-7 may need supplemental updates.

### DIFF
--- a/source/default/_patterns/00-protons/non-printing/_bootstrap-overrides.scss
+++ b/source/default/_patterns/00-protons/non-printing/_bootstrap-overrides.scss
@@ -18,17 +18,11 @@ $card-border-width: rem-calc(1px);
 
 // Spacing
 // Bootstrap Spacers use $spacer var of 1rem and assumes browser 16px default.
+// Updating the values in $spacers list adds the value to Bootstrap's utility spacers.
+// Example, the below adds a mb-6 utility spacing class.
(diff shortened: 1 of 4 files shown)


source/default/_patterns/00-protons/non-printing/_bootstrap-overrides.scss
```diff
@@ -17,9 +17,18 @@ $btn-border-radius: 0.25rem;
 $card-border-width: rem-calc(1px);
 
 // Spacing
+// Bootstrap Spacers use $spacer var of 1rem and assumes browser 16px default.
+$spacer: 1rem;
 $spacers: (
-  6: 1rem * 3.5,
-  7: 1rem * 4,
+  0: 0, // 0px
+  1: ($spacer * .25), // 4px
```
illepic: How much of this duplicates Bootstrap's defaults?

jryanconklin:
@illepic Updated! The docs are changed to reflect real usage to access the $spacers Bootstrap list via a new `spacers` function.

PR #536, opened 2019-04-16
Remove travis tests and docs for php5 and node 6

illepic:


dslemay:
@illepic We can probably safely remove the Babel polyfill from the Jest setup as well. We had to import it only for `Object.entries` support which was missing from Node v6. https://github.com/phase2/particle/blob/feature/drop-php5-node6/tools/tests/unit/setupJest.js#L1

illepic:
Nice catch! Updating now.

PR #539, opened 2019-04-18
Ignore vendor directory in eslint

jhedstrom:
Since these are used in development of particle itself, it is safe to remove them when folks install this for actual use.

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -2,6 +2,7 @@
 
 dist/
 pattern-lab/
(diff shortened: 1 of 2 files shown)


jhedstrom:
Turns out we need vendor dev dependencies as that's how particle tests itself. This now simply ignores vendor on eslint, and the fix to ignore vendor in php standards has been moved to octane itself.

PR #551, opened 2019-05-18
Add a demo block pattern and generator to ensure formatted patterns

jryanconklin:
This PR:
* Adds a new demo block proton pattern.
* Updates the generator to include the pattern demo block

Effectively this creates a containerized single column demo header with some yaml structure for demo-ing new patterns.

### DIFF
--- a/source/default/_patterns/00-protons/demo/demo-block.twig
+++ b/source/default/_patterns/00-protons/demo/demo-block.twig
@@ -1,0 +1,25 @@
+<div class="container">
+  <div class="row">
+    <div class="col">
+      {% block usage %}
+        {% if demo.title %}
+          <h1 class="mb-3">{{ demo.title|capitalize }}</h1>
+        {% endif %}
+        {% if demo.text %}
+          <div class="mb-3 p-3 bg-light border rounded">
+            {% if demo.text.lead %}
+              <p class="lead">{{ demo.text.lead }}</p>
+            {% endif %}
+            {% if demo.text.summary %}
+              <p class="summary">{{ demo.text.summary }}</p>
+            {% endif %}
+          </div>
+        {% endif %}
+      {% endblock usage %}
+
+      {% block content %}
+        <div>A simple demo block for new components.</div>
+      {% endblock content %}
+    </div>
+  </div>
+</div>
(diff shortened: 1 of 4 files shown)


tools/new-component/templates/demo-pattern.twig
```diff
@@ -1 +1,9 @@
-{% include '@<%= cleanPatternType %>/_<%= name %>.twig' %}
+{% embed '@protons/demo/demo-block.twig' with {
```
illepic: Can we use `extends` here instead to:

1. Avoid indentation and nesting
1. Indicate that this is a perfect use-case for extending

source/default/_patterns/00-protons/demo/demo-block.twig
```diff
@@ -0,0 +1,25 @@
+<div class="container">
```
illepic: I can be swayed here, but wanted to run it by you:

The only "rule" we've laid down for the difference between protons and atoms is that

>Protons are not directly usable in Twig via embed/extend/include

This will be the first proton that breaks that rule. Ideas/counter points?

tools/new-component/templates/demo-pattern.twig
illepic: Note: we can set vars in extends.

source/default/_patterns/00-protons/demo/demo-block.twig
jryanconklin: Sure. I think for me the view on protons are that they're mainly design configuration or work that supports Pattern Lab. I wouldn't include this pattern as a part of my design implementation, but it's useful in the terms of Particle to help demonstrate a use. 

Secondly, placing it in Protons means it is _never_ swept into an application as a useable pattern. I'd never want someone to use this within Drupal or Grav.

Lastly, I think it's a bad usage to include a pattern that's created later in the chain. I think the second place this makes sense is templates, but I find it problematic for an atom to include/embed/extend a template. We've tried that before and it really complicates deleting unused patterns.

illepic: **THE MOTION PASSES**

tools/new-component/templates/demo-pattern.twig
jryanconklin: Updated!

PR #552, opened 2019-05-21
Namespace writer ignores empty folders

illepic:


apps/drupal/particle.app.config.js
```diff
@@ -21,6 +21,16 @@ const APP_DIST = path.join(PATH_DIST, APP_NAME, ASSETS_BUNDLE_FOLDER);
...
+// Drupal does not use protons or pages for its namespaces. Referencing
+// folder paths that are not output to dist/ causes Drupal to kersplode.
+// const actualPaths = require('../../stats');
+// console.log(
+//   actualPaths.assets
+//     .map(asset => asset.name)
+//     .filter(name => name.includes('atomic'))
+// );
+
+const rejectNamespaces = /(00-protons|05-pages)/;
```
jhedstrom: Rather than hard-code this, wouldn't it be better to check if any twig templates actually exist in those folders before adding them to `particle.info.yml`? (As I understand it, that's the root cause of the issue--declaring them in that file but then the directories don't exist)

jryanconklin: I think this is a good check for now. Eventually it’ll be from webpack output, which is the better check here but a higher lift.

PR #569, opened 2019-06-21
Allows the pa11y domain to be overridden via a variable.

jhedstrom:
This allows `pa11y` to run in GitLab, within Docker locally, etc.

### DIFF
--- a/tools/tests/accessibility/pa11y.js
+++ b/tools/tests/accessibility/pa11y.js
@@ -1,11 +1,10 @@
 // All non-protons PL paths
 const { componentPaths } = require('../../pl-paths.js');
 
+const { PL_BASE_DOMAIN } = process.env;
+
 // The PL_BASE_DOMAIN environment variable should be set to override the default.
-let plRootDomain = '0.0.0.0:8080';
-if (process.env.PL_BASE_DOMAIN) {
-  plRootDomain = process.env.PL_BASE_DOMAIN;
-}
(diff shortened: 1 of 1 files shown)


tools/tests/accessibility/pa11y.js
```diff
@@ -1,7 +1,12 @@
 // All non-protons PL paths
 const { componentPaths } = require('../../pl-paths.js');
```
illepic:
```js
const { PL_BASE_DOMAIN } = process.env;
```

tools/tests/accessibility/pa11y.js
```diff
@@ -1,7 +1,12 @@
...
 
-const plRoot = 'http://0.0.0.0:8080/app-pl/pl'; // @TODO: move this to config
+// The PL_BASE_DOMAIN environment variable should be set to override the default.
+let plRootDomain = '0.0.0.0:8080';
+if (process.env.PL_BASE_DOMAIN) {
+  plRootDomain = process.env.PL_BASE_DOMAIN;
+}
```
illepic:
```js
const plRootDomain = PL_BASE_DOMAIN || '0.0.0.0:8080';
```

illepic: Note: still want to get this into config, per Pattern Lab app.

tools/tests/accessibility/pa11y.js
jhedstrom: Done!

PR #576, opened 2019-06-27
Add svg stroke option to svg icon template

joannasaerom:


jryanconklin:
@illepic Any thoughts here or okay to merge?

PR #598, opened 2019-08-16
Change import format of font icons to reduce bundle size

illepic:
This should drop about 40% off the final bundle size.


dslemay:
Breaking some tests on adding specific icons with a prefix is not defined error.

illepic:
@dslemay GOOD CATCH. Fix pushed, waiting for CI.

PR #612, opened 2019-09-11
Various tweaks and updates

illepic:
- Duplicate svg color output, nuke token file
- proper map-get()
- move scss2json up to pl webpack


webpack.particle.js
```diff
@@ -103,7 +104,7 @@ module.exports = {
         test: /\.(png|jpg|gif|svg)$/,
         loader: 'file-loader',
         options: {
-          name: '[name].[ext]?[hash]',
+          name: 'images/[name].[ext]?[hash]',
```
illepic: This change means that we need to update image twig paths. DO NOT MERGE until this is done.

webpack.particle.js
```diff
@@ -12,6 +12,9 @@ const { ProgressPlugin, ProvidePlugin } = require('webpack');
 const postcssPresetEnv = require('postcss-preset-env');
 const cssnano = require('cssnano');
 
+// Loaders Options
+const prefixerConfig = { autoprefixer: { grid: true } };
+
```
illepic: Idea:
```js
const postcssPresetEnvConfigured = postcssPresetEnv({ autoprefixer: { grid: true } });
```

Then down in the loader, just:
```js
? [postcssPresetEnvConfigured ]
: [postcssPresetEnvConfigured , cssnano()],
```

webpack.particle.js
illepic: Solving this today.

Issue #624, opened 2019-10-18
Migrate to Renovate Dependency Management


dslemay:
## The Problem

Currently we are using Greenkeeper to "manage" our npm dependencies. However, there is a core issue with how Greenkeeper performs this process. It only opens PRs for major version updates or updates which break a build. This can cause long gaps of dependency updates to stack up, which cause widesweeping updates. This form of update can be costly in terms of time, energy, and the possibility of inter-dependency failure from a blanket npm upgrade.

## The Solution

[Renovate](https://renovatebot.com/) is a more configurable dependency management tool. It offers a slew of benefits that are not available to us in our current Greenkeeper configuration:

* The ability to open PRs for all version bumps, not just major
* The ability to configure certain dependencies to be auto merged if all tests are passing. This is completely optional. See outstanding questions below for further discussion.
* Grouping monorepo package updates in 1 PR. For example when the Babel monorepo receives updates a PR is opened for all of the packages in the monorepo.

I have been using Renovate on several personal projects including ones that are automatically published to NPM for about the past 3 months. It is by far the best package management solution I have found. Here is what I would propose.

* Create a development branch that all work is not released is based off of.
* Configure Renovate to use the development branch as a base branch.
* This way when developers work with Particle they can receive the latest dependencies. This would require communication to expect running `npm install` when pulling develop as we would have more frequent dependency updates.
* When we are ready to publish a new version of Particle, merge into master and tag. This also keeps master pure to what is "released"

## Outstanding questions

* Are there any concerns with this as a new development approach? Any additional requirements that should be considered?
* Are there any types of dependencies we would want auto-approved? Examples could be all minor and patch devDependencies. We can also include or exclude certain dependencies with regex in the config.
* If we move forward with this approach are there any labels, reviewers, or other PR metadata we would want attached to each PR?

PR #629, opened 2019-10-25
Add Renovate Configuration

dslemay:
Related to #624 

This begins the process of adding Renovate configuration for more granular dependency management. Several questions need to be answered before being merged.

1. Do we want any dependencies to be able to be auto merged. Currently all minor and patch devDependencies which pass the test suite will be auto merged.
2. What do we want the base branch to be? In the issue I referenced keeping a develop branch and preserving master more along the lines of what is released. Do we want to move in this direction for the project overall, or should we have the base branch as master.
3. Should any reviewers or labels be set for the PRs?

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -2,14 +2,12 @@
   "extends": ["config:base"],
   "baseBranches": ["master"],
   "bumpVersion": null,
-  "packageRules": [
-    {
-      "depTypeList": ["devDependencies"],
-      "updateTypes": ["minor"],
-      "automerge": true
-    }
-  ],
+  "labels": ["dependency update"],
+  "patch": {
+    "automerge": true
+  },
   "prHourlyLimit": 5,
   "prConcurrentLimit": 10,
-  "rangeStrategy": "bump"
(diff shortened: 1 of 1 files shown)


illepic:
DEFINITELY want to do this. @jryanconklin's point about making sure this CI file is left behind for actual project implementation leads me to a bigger thing: Particle needs to move to a create-particle/generator. Doing this shouldn't be too hard (basically a copy-paste out of a "starterkit" folder or something).

For now, on 10.x we can pull in `rennovate.json` to get this replacing greenkeeper directly.

dslemay:
Regarding keeping the file, I will clarify that the file being there itself does not enable the configuration. It also requires enabling within the app. Similar to how we had a huskyrc but not husky installed. Not trying to punt, but wanted to clarify.

dslemay:
@illepic @jryanconklin Another configuration option to decide is how the package.json is updated. The current configuration will both bump the version in the package.json to see at a quick glance as well as in the lock file. If we only want the lock file updated (unless it is a major version bump) we can update that configuration.

Issue #798, opened 2020-04-30
BackstopJS is not configured properly.


eih:
### Issue
BackstopJS can't create reference images. All it can create is blank images.

### Reason
BackstopJS has some example settings but the `plRoot` variable in the settings is set to `http://0.0.0.0:8080/app-pl/pl` which is the URL for the Pattern Lab in previous versions.

### Solution
This variable should be set to `http://0.0.0.0:8080/app-node-pl/pl` which is the current URL for the Pattern Lab instance.

I've forked the repo and quickly fixed the issue. I'll create a pull request for that.

eih:
I believe this is solved with that PR has been merged now. There might be other serious issues with the BackstopJS integrations though, I'll share my findings and solution in another bug ticket and PR tomorrow.

PR #799, opened 2020-04-30
Set correct value for the `plRoot` variable

eih:
This variable is set to the old path for the Pattern Lab instance.
As this variable is appended with the component paths and used to
create URLs for BackstopJS scenarios, it causes problems. Namely
no reference could be created properly.

This fixes the problem.

Related issue: #798


eih:
Related issue is #798

illepic:
THANK YOU! Merging in for next minor release!

PR #824, opened 2020-05-29
Add accessibly tabs demo

madhaze:


source/default/_patterns/02-molecules/tabs/tabs.css
```diff
@@ -0,0 +1,7 @@
+
+.tabs__list-link:not([aria-selected="true"]):not(:hover) {
```
madhaze, line 2: I added css since I wanted it based on the aria-selected.

source/default/_patterns/02-molecules/tabs/_tabs.twig
```diff
@@ -0,0 +1,56 @@
...
+ *  - tabs.panel_title_element [string]: h2|(h3)|span Element to use for panel title.
+ *  - tabs.data [array]: array of tab objects.
+ *    - title [string]: Used for Tab title and Header inside of tab panel.
+ *    - panel [string]: Panel content.
+ */
+#}
+{% set seed = random() %}
+{% set title_element = tabs.panel_title_element|default('h3') %}
+<div class="tabs mb-4">
+  <ul class="tabs__list px-1" role="tablist" {{ tabs.label ? "aria-label='#{tabs.label}'" }}>
```
illepic, line 18: Since `tabs.label` is required, no need for an optional check. Just print `"aria-label={{ tabs.label }}"`.

source/default/_patterns/02-molecules/tabs/_tabs.twig
```diff
@@ -0,0 +1,56 @@
...
+    {% for item in tabs.data %}
+      {% set item_uid = "#{seed}-#{loop.index}" %}
```
illepic, line 20: I see you're using string interpolation here. We also have twig concatenation as well:

```twig
{% set item_uid = seed ~ '-' ~ loop.index %}
```

I've err'd on the side of twig concatentation, but I can be convinced to use interpolation instead.

source/default/_patterns/02-molecules/tabs/_tabs.twig
```diff
@@ -0,0 +1,56 @@
...
+      {% set tab_id = "tabs-tab--#{item_uid}" %}
+      {% set panel_id = "tabs-panel--#{item_uid}" %}
+      <li class="tabs__list-item inline-block" role="presentation">
+        <a
+          class="tabs__list-link inline-block py-1 px-8 text-black border border-gray-500 border-b-0 rounded-t hover:border-blue-600 hover:underline hover:text-blue-600 focus:border-blue-600 focus:outline-none focus:underline focus:text-blue-600"
+          id="{{ tab_id }}"
+          href="#{{ panel_id }}"
```
illepic, line 27: This inline link hash may be where the string interpolation gets confusing compared to the tilde concatenation.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+ */
+export function enable($context) {
+  // Get relevant elements and collections
+  const $tabsWrapper = $('.tabs', $context);
+
+  const deactivatePanels = ($panels) => {
+    $panels.each(function panelsHidden() {
+      $(this).prop('hidden', true);
+    });
+  };
```
illepic, line 44: Move this function outside `enable()`, throw a docblock on it.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+  $tabsWrapper.each(function eachTabsGroup() {
+    const $this = $(this);
```
illepic, line 47: Change `$this` to something descriptive like `$tabWrapper`.

source/default/_patterns/02-molecules/tabs/_tabs.twig
```diff
@@ -0,0 +1,56 @@
...
+{% set seed = random() %}
+{% set title_element = tabs.panel_title_element|default('h3') %}
+<div class="tabs mb-4">
+  <ul class="tabs__list px-1" role="tablist" {{ tabs.label ? "aria-label='#{tabs.label}'" }}>
+    {% for item in tabs.data %}
+      {% set item_uid = "#{seed}-#{loop.index}" %}
+      {% set tab_id = "tabs-tab--#{item_uid}" %}
+      {% set panel_id = "tabs-panel--#{item_uid}" %}
+      <li class="tabs__list-item inline-block" role="presentation">
+        <a
```
illepic, line 24: Why an `a` tag and not a `button`? Would `button` indicate action within the same page better? Note `rol="tab"` here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/Tab_Role

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+      $(`#${newPanelId}`, $context).prop('hidden', false);
+    };
+
+    // Add semantics are remove user focusability for each tab
+    $tabs.each(function tabsInit(i, tab) {
+      $(tab).attr('tabindex', '-1');
+
+      // Handle clicking of tabs for mouse users
+      $(tab).on('click', (e) => {
+        e.preventDefault();
```
illepic, line 73: If we use `button`s, we don't have to `e.preventDefault(0`.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+      // Hide all panels.
+      deactivatePanels($panels);
+      // Unhide new panel.
+      $(`#${newPanelId}`, $context).prop('hidden', false);
+    };
+
+    // Add semantics are remove user focusability for each tab
+    $tabs.each(function tabsInit(i, tab) {
+      $(tab).attr('tabindex', '-1');
```
illepic, line 69: `tab` is already a jQuery object here. I think you're double-wrapping by `$(tab)`. Just send it in to your function as `$tab` and use it directly.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+      // Handle clicking of tabs for mouse users
+      $(tab).on('click', (e) => {
+        e.preventDefault();
+        const $currentTab = $('[aria-selected="true"]', $tablist);
+        // Switch tab if clicked tab is not current tab.
+        if ($(e.currentTarget)[0] !== $currentTab[0]) {
```
illepic, line 76: What's with array accessor here (`[0]`)? There's a better way to do this here.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+      });
+
+      // Handle keydown events for keyboard users
+      $(tab).on('keydown', (e) => {
+        // Get the index of the current tab in the tabs node list
+        const index = $tabs.index(e.currentTarget);
+
+        // Work out which key the user is pressing and
+        // Calculate the new tab's index where appropriate
+        const { key } = e;
```
illepic, line 88: Niiiiice.

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+        if (key === 'ArrowLeft') {
+          e.preventDefault();
```
illepic, line 90: Do we have to `preventDefault` on keypresses?

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+        }
+      });
+    });
+
+    // Add tab panel semantics and hide them all.
+    $panels.each((i, panel) => {
+      $(panel).attr('tabindex', '-1');
+      $(panel).attr('aria-labelledby', $tabs[i].id);
+      $(panel).attr('hidden', true);
+    });
```
illepic, line 116: Isn't this redundant since you're already setting these defaults in twig?

source/default/_patterns/02-molecules/tabs/index.js
```diff
@@ -0,0 +1,125 @@
...
+
+    // Initially activate the first tab and reveal the first tab panel.
+    $($tabs[0]).removeAttr('tabindex');
+    $($tabs[0]).attr('aria-selected', 'true');
+    $($panels[0]).attr('hidden', false);
```
illepic, line 121: Use `.first()`: `$tabs.first().removeAttr()`. Anyway, couldn't you set this in twig to start out?

source/default/_patterns/02-molecules/tabs/tabs.css
illepic: Comment both these lines of CSS with what's going on here.

illepic:
Also, please pull in `master` and update this branch with those conflicts resolved.

source/default/_patterns/02-molecules/tabs/_tabs.twig
madhaze: I think I was doing interpolation since in our eslinting of js we must use interpolation. Tho the syntax is more finicky than concatenation.

source/default/_patterns/02-molecules/tabs/_tabs.twig
madhaze: twig doesn't allow us to specify required variables so shouldn't this be a check

illepic:
I have changes I'm working on in this file.

source/default/_patterns/02-molecules/tabs/_tabs.twig
```diff
@@ -0,0 +1,53 @@
...
+        </a>
+      </li>
+    {% endfor %}
+  </ul>
+
+  <div class="tabs__panels">
+    {% for item in tabs.data %}
+      {% set item_uid = "#{seed}-#{loop.index0}" %}
+      {# A11y show initially let js choose to which to show/hide. #}
+      <section class="tabs__panels-item border border-gray-500 p-4 bg-white focus:border-blue-600 focus:outline-none"
```
jryanconklin, line 39: Is it worth placing these attributes into an array and allowing `attributify` filter handle these?

source/default/_patterns/02-molecules/tabs/__tests__/tabs.test.js
```diff
@@ -0,0 +1,144 @@
+import $ from 'jquery';
+import tabs, { name, selectedIndexState, updateSelectedState } from '..';
+
+const markup = `
```
jryanconklin, line 4: Is this necessarily the pattern we want to use for testing markup generated by our patterns? Since test coverage of default patterns is minimal, this will become the defacto procedure for developers to follow. Would it be possible to capture the generated content from Pattern Lab and test against that markup? If so, does it then make more sense to move our tests into demos?

PR #826, opened 2020-06-01
Update broken image demos.

madhaze:


source/default/_patterns/01-atoms/image/_image--bg.twig
```diff
@@ -9,4 +9,4 @@
  *  - image.repeat [string]: Background repeat value for image.
  */
 #}
-<div style="background-image: url({{ image.src }}); background-repeat: {{ image.repeat|default('no-repeat') }}" class="{{ image_classes }}"></div>
+<div style="background-image: url({{ image.src }}); background-repeat: {{ image.repeat|default('no-repeat') }}" class="{{ image_classes }}" role="img" aria-label="{{ image.alt }}"></div>
```
madhaze, line 12: Added aria-label using alt text.  Tho with Object-fit being supported (well not in IE11) Maybe we should switch to that over this way of handling background images.

https://caniuse.com/#feat=object-fit
https://developer.mozilla.org/en-US/docs/Web/CSS/object-fit
https://developer.mozilla.org/en-US/docs/Web/CSS/object-position

PR #829, opened 2020-06-02
SVG with Tailwind.

illepic:
SVG sprite handling update


source/default/_patterns/01-atoms/svg/sprite-svg.css
```diff
@@ -0,0 +1,5 @@
+[class^='sprite-'] {
+  @apply w-6 h-6;
+
+  display: inline-block;
```
dslemay, line 4: This could use the Tailwind class for inline block as well.

```suggestion
  @apply w-6 h-6 inline-block;
```

illepic: flerp

apps/node-pl/pattern-lab/_patterns/01-atoms-demo/svg/svgs.twig
```diff
@@ -2,9 +2,106 @@
...
+			</div>
+      <!--
+        Embed: Leveraging additional classes being passed directly to the SVG
+        component, no need for an extra wrapping <div>
+      -->
+      {% include '@atoms/svg/_svg.twig' with {
+        svg: {
+          path: '@atoms/svg/icons/search.svg',
+        },
+        svg_classes: ['text-red-500', 'w-4'],
```
jryanconklin, line 41: Colors are a really likely point of Tailwind config we'll remove. I've tried to only use `gray` in docs with the thought that once we remove colors this is a point of breakage. We might consider skipping colors in demo.

apps/node-pl/pattern-lab/_patterns/01-atoms-demo/svg/svgs.yml
```diff
@@ -1,8 +1,7 @@
 svg_demo_block:
   title: 'SVG Component'
-  text:
+  content:
```
jryanconklin, line 3: My bad, thank you for fixing!

apps/node-pl/pattern-lab/_patterns/01-atoms-demo/svg/svgs.twig
illepic: "Before removing colors in tailwind, ensure you command+f and search for any of the removed colors, ie `red-500`"

source/default/_patterns/01-atoms/svg/_svg.twig
```diff
@@ -3,29 +3,43 @@
...
+} | merge(svg) %}
 
-{# Classes Merge Handled by Variant Files #}
-{% set svg_classes = svg_classes|default([])|merge([
+{# Default SVG classes, overridden by incoming array #}
+{% set svg_classes = [
   'svg',
-  svg_variant,
-]) %}
+  'svg--' ~ svg.variant,
```
madhaze, line 27: With purgeCss do we need to update this to be full classes. 
`
```suggestion
{% set variants = {
    embed: ['svg--embed'],
    sprite: ['svg--sprite'],
} %}

{% set svg_classes = [
  'svg',
  svg.fillCurrent ? 'fill-current' : '',
  svg.strokeCurrent ? 'stroke-current' : '',
] 
  | merge(attribute(variants, svg.variant))
  | merge(svg_classes|default([]))
  | filter(c => c is not empty) | join(' ') %}

```

madhaze: ok, that suggestion was for not just replacing that one line. but lines 25-30

jryanconklin: @madhaze I don't think that will be completely necessary now that Tailwind implements PurgeCSS natively. It only purges itself.

source/default/_patterns/01-atoms/svg/sprite-svg.css
```diff
@@ -0,0 +1,3 @@
+[class^='sprite-'] {
+  @apply w-6 h-6 inline-block;
```
madhaze, line 2: This does not allow for changing the size of an icon when used in dom. Like `<i class="sprite-close w-2"></i>` will not affect the width. The ^^ css has a higher specificity than `w-2`.

source/default/_patterns/01-atoms/svg/_svg.twig
madhaze: so we are only purgecssing tailwind classes?

jryanconklin: Correct! It's very cool!

source/default/_patterns/01-atoms/svg/sprite-svg.css
illepic: Noted. Looking at a solution here now.

illepic: Check this file and the demo now. Should be resolved.

PR #830, opened 2020-06-03
Accordion component

madhaze:


source/default/_patterns/02-molecules/accordion/_accordion.twig
```diff
@@ -0,0 +1,49 @@
...
+{% set seed = random() %}
+{% set trigger_element = accordion.trigger_element|default('h3') %}
+<div class="accordion mb-4 focus-within:shadow-outline"
+  {{ accordion.allow_toggle or accordion.collapsed ? 'data-allow-toggle="true"' }}
+  {{ accordion.allow_multiple ? 'data-allow-multiple="true"' }}
+>
+  {% for item in accordion.data %}
+    {% set item_uid = "#{seed}-#{loop.index}" %}
+    <{{ trigger_element }}>
+      <button {{ loop.first and not accordion.collapsed ? 'aria-expanded="true"' }}
```
jryanconklin, line 24: Can I talk you into using an attributes array and the `attributify` filter? We could use some better examples of it working in Particle and it would also allow us to update any amount of attributes in markup.

Issue #827, opened 2020-06-02
Update/Add Documentation on how to use Particle 10.8.0


Jwaxo:
Installing 10.8.0 this morning, there are definitely some missing gaps in basic knowledge transfer. Installation/quickstart guides are still in place, but, for example, when trying to figure out how to use the new setup for the Drupal app I needed to have a few questions answered face-to-face, when it seems they could be documented in text. I may take this task upon myself to at least add some barebones READMEs scattered throughout the new folder structures, but thought I'd at least make an issue for it to track.

bmartinez287:
So, it be great if the documentation could include examples of using both tailwind and scss.

bmartinez287:
Could we get an update on this page?
https://phase2.gitbook.io/frontend/particle/architecture/component-structure
With the addition of tailwinds, it seems like webpack does not like scss anymore.

PR #831, opened 2020-06-03
TASK :: Revise Drupal + Pattern Lab Directory Structure

jryanconklin:
# Pull Request

Closes Issue #827 

## Issue Description

Revise directory structure and docs for easier understanding.

## Summary of Changes

Revise the directory structure and update Drupal documentation. This PR:
* Update `apps/drupal` => `apps/drupal-default`to match `thing-designsystem` naming
* Update Drupal Theme Path and rename `apps/drupal-default/particle_theme`
* Update Drupal Module and rename `apps/drupal-default/particle_helper`
* Update `apps/node-pl` => `apps/node-pl-default`to match `thing-designsystem` naming


mike-potter:
I feel like `node-pl-default` should just be `pl-default`.  Don't see the need for the `node` distinction anymore.  Also not generally a fan of `-default` everywhere.  You could easily have a convention of `thing-designsystem` where if `designsystem` was not specified it would use `default` as the default.  

Don't hugely care as long as the URLs dont end up being even longer for getting to PL from Drupal `/dist/apps/node-pl-default/pl` or something.

Finally, in actual Drupal sites using Octane, the theme will be in `/project/themes/THEMENAME` and the PL design system (Particle) will be in `/project/design/particle` (until Particle moves to being an actual node dependency, at which point the "source" patterns would be in `project/design/DESIGNNAME`.  So just want to ensure Particle config will still allow that kind of directory structure.

illepic:
If we're in here making these changes (finally!), let's drop "node" off the app name.

@mike-potter, we're looking for a quick naming convention to tie an "app" to a "design system", since a drupal theme and a pattern lab (and, eventually, a storybook) consume and represent exactly one design system. We're just looking for a quick glance way of knowing "what is this app showing off?" to make the eventual particle 11 generators work.

Those generators will effectively ask us these questions in this order:

1. What do you want to name your design system?
1. Where should it be placed?
1. Which apps (drupal|pl) would you like to consume this design system?
1. Where should those apps live?

illepic:
Whoops, hit tab and accidentally closed! My bad!

jryanconklin:
> I feel like `node-pl-default` should just be `pl-default`. Don't see the need for the `node` distinction anymore. Also not generally a fan of `-default` everywhere. You could easily have a convention of `thing-designsystem` where if `designsystem` was not specified it would use `default` as the default.
> 
> Don't hugely care as long as the URLs dont end up being even longer for getting to PL from Drupal `/dist/apps/node-pl-default/pl` or something.
> 
> Finally, in actual Drupal sites using Octane, the theme will be in `/project/themes/THEMENAME` and the PL design system (Particle) will be in `/project/design/particle` (until Particle moves to being an actual node dependency, at which point the "source" patterns would be in `project/design/DESIGNNAME`. So just want to ensure Particle config will still allow that kind of directory structure.

Default isn't really meant to be kept and I actually like it as a paint point to rename your DS from `default`. But moreover this will clear the way to setting a DS name as config in 11. I think `default` is fine as a starting point but it can as easily be `base` (though I question the labor involved to make that change on 10). TLDR: this will be useful for generators and just basic cognitive understanding of what apps are doing.

I don't think the docs should count on Octane project structure. That tooling has everything it needs to know where to put Particle. Whereas the docs need to cover how to use this elsewhere.

jryanconklin:
@illepic I'll update to remove `node` from `apps/node-pl-default`

jryanconklin:
@illepic @mike-potter Updated PR:
* `apps/node-pl-default` => `apps/pl-default`
* Config for `app-pl-node` => `app-pl`

mike-potter:
OK, that looks good.

I wasn't asking for a change in docs, I just wanted to make sure that the directory structure needed for Octane can be supported.  Sounds like the answer is "yes".  And I am fine with `default` being a "pain point" to encourage projects to name their design system and theme correctly.

@Bloom so in your questions, Octane would answer:

What do you want to name your design system (default: "default")? `octane`
Where should it be placed (default: source/default)? `project/design/octane`
Which apps (drupal|pl) would you like to consume this design system (default: both selected)? `both`
Where should those apps live (default: apps/)?
*this question needs to be asked for each app*
drupal: `project/themes/octane`
pl: `project/apps/pl` (or something like that...we don't currently use PL outside of the DESIGN_PATH but supporting this same app concept would be useful)

illepic:
This no longer run locally via `npm start` (even after an `npm run setup`). I'm looking into why now.

jryanconklin:
Boo. Okay. I can take a look, too. Did you pull without doing a full reset? I have been nuking everything between steps and verifying with ‘npm run ci’

jryanconklin:
@illepic, confirmed on my local machine. Are you deleting node modules and removing old dist?

illepic:
NOPE. That's my machine due to some port mappings I had to do recently for Docksal. Carry on! I'll get this figured out.

PR #858, opened 2020-06-20
PHP deprecated warning: Swapping parameters order in implode() function.

yuginaili:


apps/pl-default/pattern-lab/alter-twig.php
```diff
@@ -192,7 +192,7 @@ function addFilters(\Twig_Environment &$env, $config) {
         }
         $renderable_attributes[] = "{$key}=\"{$value}\"";
       }
-      return implode($renderable_attributes, ' ');
+      return implode(' ', $renderable_attributes);
```
jryanconklin, line 195: @yuginaili `implode` is also used above on line 191.

yuginaili: Thanks @jryanconklin

PR #892, opened 2020-07-13
Adding shareable eslint rules

mealeyst:
# Adding shareable eslint rules

## Issue Description

This PR attempts to set up the particle codebase to utilize shareable eslint configs.

To use the configs one should only need to do the following for their `.eslintrc.js`

### Base eslint rules
```
module.exports = {
  extends: ['@phase2/eslint-config']
}
```

### React eslint rules
```
module.exports = {
  extends: ['@phase2/eslint-config/react', '@phase2/eslint-config/typescript']
}
```

## Summary of Changes

Split up the eslint files to be used individually or combined together

### DIFF
--- a/packages/eslint-config/jest.js
+++ b/packages/eslint-config/jest.js
@@ -1,0 +1,6 @@
+module.exports = {
+  extends: ['plugin:jest/recommended', './index.js'],
+  rules: {
+    'jest/expect-expect': 'off',
+  },
+}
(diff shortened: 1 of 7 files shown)


packages/eslint-config/jest.js
```diff
@@ -0,0 +1,6 @@
+module.exports = {
+  extends: ['plugin:jest/recommended', './index.js'],
+  rules: {
+    'jest/expect-expect': 'off',
```
FrederickEngelhardt, line 4: Do we need this disabled?

PR #904, opened 2020-07-21
Move generators to top level packages

FrederickEngelhardt:
Quick PR to move all generators up to top level packages.

Future TODOs to pull out inquirer prompt into each sub package


mealeyst:
This looks good to me based on the pair programming that you and I did. I am going to see if I can get the Pattern Lab generator to follow this as well.

Issue #905, opened 2020-07-22
`npm run new` fails


illepic:
`npm run new` is erroring out.

PR #909, opened 2020-07-24
Fix `npm run new`

illepic:
# Pull Request

Closes Issue https://github.com/phase2/particle/issues/905

## Issue Description

`npm run new` was failing to write out files with proper extensions. Also updates deps.

## Summary of Changes

Yeoman released a fix to internal file writing mechanisms that invalidated a stopgap measure we had in place.


tools/generators/new-component/generators/app/index.js
```diff
@@ -148,12 +148,9 @@ module.exports = class extends Generator {
     // a reserved method to which Yeoman provides all file streams from copyTpl()
     this.registerTransformStream(
       rename((path) => {
-        // basename is 'patterns.twig' here
-        const ext = extname(path.basename);
-        // Original extname was '.ejs', change it to ext, which is now '.twig'
-        path.extname = ext;
-        // Remove extension ('.twig') from basename, replace 'pattern" with name
-        path.basename = path.basename.replace(ext, '').replace('pattern', name);
+        // basename is ie '_patterns.twig' here
```
illepic, line 151: All the stuff above here was a duct tape fix due to a quirk in Yeoman. Looks like Yeoman fixed the quirk.

adamjuran:
Possibly unrelated, we've had issues when yoeman is installed globally... The end result being that a new component is created in the directory/project where the theme was last installed. Does this fix that or is that a whole new issue?

illepic:
@adamjuran [this file](https://github.com/phase2/particle/blob/master/.yo-rc.json) should indicate to Yeoman the "root" from which to run. `npm run new` should be reaching for locally-installed yeoman first as well. I'd be super interested in your experience. If you can replicated it, hit me with a ticket and I'll tackle. We're going to use yeoman a lot for the project generator in Particle 11, so this is great to know!

Issue #919, opened 2020-07-28
Feature: Build config file


mealeyst:
Build `.particle-rc` based on user input from prompts and include cli version

PR #924, opened 2020-07-29
Saving options in .particle-rc

mealeyst:
# Pull Request

Closes Issue #919

## Issue Description

This PR adds saves the user's inputs in an `.particle-rc` file that includes the configuration options that the user has selected as well as the version of the cli that the user executed the command from.

## Summary of Changes

After the user is prompted for answers, we take those answers as well as the cli version and save it into a `.particle-rc` file.

### DIFF
--- a/packages/generator-particle-base/src/generators/app/index.ts
+++ b/packages/generator-particle-base/src/generators/app/index.ts
@@ -19,12 +19,10 @@ module.exports = class extends Generator {
   // configuration will come from the constructor argument
   configuration: Answers
   packageJson: Record<string, any>
-  cliVersion: string
+  cliVersion = ''
 
   constructor(args: any, opts: any) {
     super(args, opts)
-    console.log(args)
-    this.cliVersion = ''
     for (let i = 0; i < args.length; i++) {
       const value = args[i]
       const regex = /(cli-version=)(\d.*)/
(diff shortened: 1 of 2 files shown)


packages/generator-particle-base/src/generators/app/index.ts
```diff
@@ -82,7 +93,14 @@ module.exports = class extends Generator {
         options: configOptions[results.config],
       }
     }
-
+    fs.writeFileSync(
```
FrederickEngelhardt, line 96: Can we move this fs call outside of the constructor. Feels cleaner to have a method called `writeParticleJson()`.

packages/generator-particle-base/src/generators/app/index.ts
```diff
@@ -19,9 +19,20 @@ module.exports = class extends Generator {
   // configuration will come from the constructor argument
   configuration: Answers
   packageJson: Record<string, any>
+  cliVersion: string
```
FrederickEngelhardt, line 22: You can make this an empty string and move 27 out of the constructor.

packages/generator-particle-base/src/generators/app/index.ts
```diff
@@ -19,9 +19,20 @@ module.exports = class extends Generator {
...
 
   constructor(args: any, opts: any) {
     super(args, opts)
+    console.log(args)
+    this.cliVersion = ''
+    for (let i = 0; i < args.length; i++) {
+      const value = args[i]
```
FrederickEngelhardt, line 29: Do we want a console log here?

PR #942, opened 2020-08-18
eslint-setup: Adds logic to create simplified eslintrc.js file.

LeeMellon:
# Pull Request

Adds logic to create simplified eslintrc.js file.
nukeme____dev_nukeme__-______eslintrc_js_and_iangoodrich_elfayoumi____dev_particle

### DIFF
--- a/packages/generator-particle-base/src/generators/app/index.ts
+++ b/packages/generator-particle-base/src/generators/app/index.ts
@@ -1,6 +1,7 @@
 import Generator from 'yeoman-generator'
 import merge from 'lodash.merge'
 import fs from 'fs'
+import _ from 'lodash';
 
 import {
   Answers,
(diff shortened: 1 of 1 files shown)


packages/generator-particle-base/src/generators/app/index.ts
```diff
@@ -88,6 +89,26 @@ module.exports = class extends Generator {
     )
   }
 
+  /**
+   * Create eslintrc file
+   **/
+
+  _eslintCompile(configuration: Answers ) {
+    const {frontendFramework, testingLibraries, hasTypescript } = configuration.options;
```
mealeyst, line 97:
```suggestion
    const { frontendFramework, testingLibraries, hasTypescript } = configuration.options;
```

PR #951, opened 2020-09-06
Revise SVG Handling

jryanconklin:
# Pull Request

Closes Issue #NO-TICKET

## Issue Description

Simplify SVG use and add better Tailwind SVGO config. 

## Summary of Changes

Revise SVG handling to move file directory to static directory. Update webpack to add directly to  `assets/images` for simpler bundle structure. Add `@static` Pattern Lab namespace to coerce PL to use pre-compiled assets directly from file system. Update SVGO config to ensure `fill='currentColor'` to mimic Tailwind SVG handling (heroicons project).


.svgo.yml
```diff
@@ -19,7 +19,11 @@ plugins:
   - minifyStyles: true
   - moveElemsAttrsToGroup: true
   - moveGroupAttrsToElems: true
-  - removeAttrs: false
+  - removeAttrs:
+      attrs: 'fill'
+  - addAttributesToSVGElement:
+      attributes:
+        - fill: 'currentColor'
```
illepic, line 26: Can we solve this with `svg { @apply fill-current; }` instead?

source/default/webpack.config.js
```diff
@@ -31,27 +28,17 @@ module.exports = {
...
     // Sprite system options
     new SVGSpritemapPlugin(
-      path.resolve(namespaces.atoms, 'svg/icons/**/*.svg'),
+      path.resolve(namespaces.static, 'icons/svg/**/*.svg'),
       {
         output: {
           filename: 'images/spritemap.svg',
           svg4everybody: true,
-          svgo: true,
+          svgo: false,
```
illepic, line 41: Why disable this?

illepic (changes requested):

source/default/webpack.config.js
jryanconklin: @illepic Because we know we want to run these manually based on our own SVGO configuration. We've already cleaned these by the time this plugin is involved and we don't want to remove currentColor.

.svgo.yml
jryanconklin: Ah! It’s not entirely the same in our setup. Because we're wrapping the svg markup in span, the svg itself never gets the attribute value in some situations and for icons we near always want the currentColor on fill. I borrowed this approach from Tailwind's own SVG library: https://github.com/tailwindlabs/heroicons/blob/master/svgo.solid.yaml

Secondarily, this value is how the Tailwind SVG Plugin I wrote works. When a file is scanned and currentColor exists on fill or stroke, you replace it with the color vars you apply in the plugin.

PR #955, opened 2020-10-26
10.8

illepic:


.eslintrc.js
```diff
@@ -6,19 +6,12 @@
...
-    'prettier/vue',
-  ],
-  plugins: ['prettier'],
+  extends: ['airbnb-base', 'plugin:jest/recommended', 'prettier'],
   root: true,
   globals: {
     Drupal: true,
     jQuery: true,
-    _: true,
+    // _: true,
```
jryanconklin, line 14: Is this intentional?

apps/drupal-default/webpack.config.js
```diff
@@ -3,31 +3,23 @@
...
+const drupalWebpackConfig = {
   entry: {
     'drupal-jquery': [path.resolve(__dirname, 'drupal-jquery.js')],
     app: [path.resolve(__dirname, 'index.js')],
   },
   output: {
-    path: APP_DIST,
-    publicPath: APP_DIST_PUBLIC,
+    // Output all CSS/JS/images/twig to dist/ within drupal theme
+    path: path.resolve(__dirname, 'particle_theme/dist'),
```
jryanconklin, line 22: Octane programmatically moves this directory.

apps/pl-default/webpack.config.js
```diff
@@ -2,35 +2,39 @@
...
 
 // Plugins
 const RunScriptOnFiletypeChange = require('../../tools/webpack/run-script-on-filetype-change');
-const particle = require('../../particle');
 
 // Constants: environment
 const { NODE_ENV, PARTICLE_PL_HOST = '' } = process.env;
 // Constants: root
-const { PATTERN_LAB_DIST } = require('../../particle.root.config');
+const { PATH_DIST } = require('../../particle.root.config');
```
jryanconklin, line 16: Are we undoing having particle split out from the Drupal theme?

postcss.config.js
```diff
@@ -2,26 +2,29 @@
...
 const cssnano = require('cssnano');
-const tailwindcss = require('tailwindcss');
 const hexrgba = require('postcss-hexrgba');
-const stylelint = require('stylelint');
-const postcssReporter = require('postcss-reporter');
 
-module.exports = ({ options, env }) => {
+const tailWindPath = path.resolve(
+  __dirname,
+  './source/default/tailwind.config.js'
```
jryanconklin, line 14: This might be nice to set up as config since default may not survive as a name.

source/default/tailwind.config.js
```diff
@@ -19,23 +19,25 @@ module.exports = {
...
+    // Replace default Tailwind config here
     colors,
     customForms,
     fontFamily,
+    // Extend (add to) default Tailwdind config here
     extend: {},
   },
   variants: {},
   plugins: [tailwindCustomForms],
+  future: {
```
jryanconklin, line 37: Might be worth a comment here that future can be removed in Tailwind 2.0

jryanconklin (changes requested):
There’s an empty spacing.json file. Beyond that it’s worth noting this approach will break Octane scripts, which now is configured to separate the theme, module, and particle into separate directories. Considering the priority of the Octane project, I wouldn’t expect this to be changed until the New Year.

PR #956, opened 2020-10-27
Eleven cli

LeeMellon:
# Pull Request

## Summary of Changes
Updates/cleans up no longer used particle-types and prompt options.
Generates .particlerc with designThemes prop.
iangoodrich_elfayoumi____dev_particle_packages_particle-cli_and_particle_–__particlerc

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,9 +46,9 @@ To remove package-lock.json from all levels of the repo simply run this command.
 ps -ef | (grep -q -s -R ^$1 package-lock.json && rm -rf package-lock.json) | { grep -v grep || true; }; lerna exec -- ps -ef | (grep -q -s -R ^$1 package-lock.json && rm -rf package-lock.json) | { grep -v grep || true; }
 ```
 
-To remove all typescript lib files run `npm dev:clean:lib`
+To remove all typescript lib files run `npm run dev:clean:lib`
 
-To remove all node_modules in packages run `npm dev:clean:node_modules`
+To remove all node_modules in packages run `npm run dev:clean:node_modules`
 
 ### Upgrading dependencies
 
--- a/packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
+++ b/packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
@@ -12,7 +12,7 @@ const minMaxOptionsValidate = ({ min, max }: { min: number; max?: number }) => (
 ) => {
   if (answer.length < min || (!max ? false : answer.length > max)) {
     return `You must choose a minimum of ${min} option(s)${
-      max ? ` and a maximum of ${max} option(s)` : ''
+      max ? ` and a maximum of ${max} option(s).` : ''
     }`
   }
   return true
@@ -37,7 +37,7 @@ export const designThemePrompt = () => [
   },
   {
     type: 'input',
-    message: 'choose a design theme name using kebab case. (min 4 chars) Ex: "alpha"',
+    message: 'Choose a design theme name using kebab case. (min 4 chars) Ex: "alpha".',
     name: 'themeName',
     default: 'default',
     validate: (name: string) => {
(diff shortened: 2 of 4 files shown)


README.md
```diff
@@ -46,9 +46,9 @@ To remove package-lock.json from all levels of the repo simply run this command.
 ps -ef | (grep -q -s -R ^$1 package-lock.json && rm -rf package-lock.json) | { grep -v grep || true; }; lerna exec -- ps -ef | (grep -q -s -R ^$1 package-lock.json && rm -rf package-lock.json) | { grep -v grep || true; }
 ```
 
-To remove all typescript lib files run `npm dev:clean:lib`
+To remove all typescript lib files run `npm run dev:clean:lib`
 
-To remove all node_modules in packages run `npm dev:clean:node_modules`
+To remove all node_modules in packages run `npm run dev:clean:node_modules`
```
FrederickEngelhardt, line 51: Good catch.

packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
```diff
@@ -2,54 +2,44 @@ import inquirer from 'inquirer'
...
+  DesignTheme,
 } from '@phase2/particle-types'
 
 const minMaxOptionsValidate = ({ min, max }: { min: number; max?: number }) => (
   answer: Record<string, string>[]
 ) => {
   if (answer.length < min || (!max ? false : answer.length > max)) {
     return `You must choose a minimum of ${min} option(s)${
-      max ? ` and a maximum of ${max} option(s)` : ''
+      max ? ` and a maximum of ${max} option(s).` : ''
```
FrederickEngelhardt, line 15: let's move this tern above the return into a variable.

```
const maxText = ` max ? ` and a maximum of ${max} option(s).` : ''`
return `You must choose a minimum of ${min} option(s)${maxText}`
```

FrederickEngelhardt: Should be a bit cleaner ^

packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
```diff
@@ -58,129 +48,88 @@ export const configurationPrompt = [
...
+  message: 'Do you want to add another theme?',
+  default: true
+}]
+
+export const generatorLoop = async() => {
+  let loop = true;
+  let themesArray:DesignTheme[] = []
+  do {
+      themesArray = await compileThemes(themesArray)
+      loop = await inquirer.prompt(rerun).then(answers => answers.generateTheme)
```
FrederickEngelhardt, line 79: Why do we need a `Promise.resolve().then()` for await syntax? 

Also not sure what answers.generateTheme does. Can you explain it? Did a code search and it looks like it just sets the loop equal to a string.

Thinking something like this
```
answers = await inquirer.prompt(rerun)
loop = false
```

packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
```diff
@@ -58,129 +48,88 @@ export const configurationPrompt = [
...
+    message: 'Choose a abbreviation for your/client\'s name. (min 3 chars)',
+    name: 'clientAbbreviation',
     validate: (name: string) => {
-      if (!name || name.length < 4) {
-        return 'Please enter a library name of more than 4 characters length'
+      if (!name || name.length < 3) {
+        return 'Please enter a project name of more than 4 characters length.'
+      }
+      if (name.indexOf(' ') > 0) {
+        return 'Please enter a two word project name with no spaces.'
```
FrederickEngelhardt, line 95: We're repeating this line below and the logic does not match the returned message. It should be `name.length < 4`. Can we create a logic block that reuses this.

example return { isValid: boolean, message: string}
```
const validateNameLength = (name:string, minLength: number) => {
 if (name && name.length >= minLength){
  return { isValid: true }
}
return { isValid: false, message:  `Please enter a project name of more than ${minLength} characters length.`}
}
```

FrederickEngelhardt: See 105 for the repeat logic.

FrederickEngelhardt: We may just be able to reuse the entire validation function. Up to you. Just want to clean up duplicates.

packages/generator-particle-base/src/generators/app/index.ts
```diff
@@ -115,20 +101,20 @@ module.exports = class extends Generator {
...
+    //   )
+    // ) {
+    //   this.composeWith(
+    //     require.resolve('@phase2/generator-particle-storybook'),
+    //     {
+    //       configuration: this.configuration,
+    //       updatePackageJson: this._updatePackageJson,
+    //     }
+    //   )
+    // }
```
FrederickEngelhardt, line 117: Is this commented out because it will be used later? Is there a need to comment it out?

packages/generator-particle-base/src/generators/app/generatePromptOptions.ts
LeeMellon: It sets loop to a bool based on user response. Doing it this way because if I include that question in  `compileThemes ` I'd have to clean it out of every object in the themes array afterwards. 
The resolution of the promise is the then. w/o that I'd be setting the loop to a pending promise resolution.

packages/generator-particle-base/src/generators/app/index.ts
LeeMellon: Honestly I'm not using it right now, but it looked useful to keep as documentation for the rest of the sub-generators to come.

